PR #8, opened 2022-07-07
Faster Usernameify

xeoncross:
There is often little reason to use regex in Go with string parsing since iterating over bytes or runes is built into the language. I saw the inflector package and realized the code could be much faster if operating directly on the underlying string. Turns out, it's about twice as fast and uses less than half the memory, there is still room for improvement.

```
BenchmarkUsernamify-old   	  530571	      2108 ns/op	     836 B/op	      15 allocs/op
BenchmarkUsernamify-new   	 1209289	      1140 ns/op	     280 B/op	      31 allocs/op
```

Please note, this is a change in API from the original `inflector.Usernamify` which returned unexpected magic constants ("unknown") instead of empty string for invalid input. Furthermore, the entire unicode letter space is now supported for usernames. This might not be desired, but is an improvement for global projects.

### DIFF
--- a/tools/inflector/inflector.go
+++ b/tools/inflector/inflector.go
@@ -93,26 +93,19 @@ func Snakecase(str string) string {
 // Usernamify("John Doe,   hello") // "john.doe.hello"
 // ```
 func Usernamify(str string) string {
-	// split at any non word character
-	words := usernamifySplitRegex.Split(strings.ToLower(str), -1)
-
-	// concatenate any non empty word with a dot
-	var result strings.Builder
-	for _, word := range words {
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
Thanks for the PR, but this actually may not be needed because the method is no longer used (it was created initially to handle admin usernames, but I ended up using the email as auth identifier as it is more generic and common).

In any case, `inflector.Usernamify` has a very specific Presentator implementation so it's better to be deleted from PocketBase.
I'll review the other methods and I'll clean up the API before people start using the package more extensively.

PR #13, opened 2022-07-08
Fix typo in PageTokenOptions

brams-dev:
previosly -> previously


ganigeorgiev:
Thanks!

_p.s. The build was manually cancelled because currently the setup requires to have the latest ui spa precompiled before running the action (the ui/dist artifacts are compared to the one generated during the runner; this weekend I'll have more time and will see if the process could be improved)._

PR #39, opened 2022-07-09
add more lint with gofumpt and replaced mutex map with sync.Map for better performance.

ValleyZw:
- added gofumpt and enabled extra lint.
- replaced mutex map with sync.Map for better performance.


ganigeorgiev:
@ValleyZw Can you rebase/pull the latest changes from master?

ganigeorgiev:
Also, please lets stick to gofmt - it is already strict enough in my opinion.

ganigeorgiev:
@ValleyZw After the reading the documentation for sync.Map it actually not the same as using a RW mutex, because it doesn't seems to be concurrent safe when multiple routines execute read_map->modify->update_map operations.
I also don't like that it makes the code a little more difficult to work with.
Unless there is something I'm missing and the eventual performance benefits are really worthed, I'd prefer to keep the current implementation.

PR #47, opened 2022-07-10
fixed some doc and code inconsistencies and removed some redundant parentheses.

ValleyZw:
- fixed some doc latencies.
- removed some redundant parentheses.
- replaced the variable `new` (collides with the 'builtin' function) with `settings`.


ganigeorgiev:
Thanks!

PR #53, opened 2022-07-10
fix migrations register in framework mode

egor-romanov:
## What kind of change does this PR introduce?

Bug fix:

Migration won't be registered if you don't provide `optFilename` currently in framework mode when you create migration.

## What is the current behavior?

Path would be `1640988000_init.go` for all migration files you try to register.

## What is the new behavior?

Now we add optFilename in `Register` function in `1640988000_init.go` if it is not provided in migration itself.

### DIFF
--- a/migrations/1640988000_init.go
+++ b/migrations/1640988000_init.go
@@ -3,6 +3,8 @@ package migrations
 
 import (
 	"fmt"
+	"path/filepath"
+	"runtime"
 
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Thanks! I've totally missed that by nesting I was changing the call stack.
I may consider searching for a more "stable" solution in the future than relying on `runtime.Caller`.

PR #60, opened 2022-07-10
fix list api docs example

clysto:
```js
client.Records.getFullList("collection", 200 /* batch size */); // <-- remove this semicolon
    .then(function (records) {
        // success...
    }).catch(function (error) {
        // error...
    });
```


ganigeorgiev:
Thanks!

_This will be available with the next minor release next week (v0.2.3) since I just tagged v0.2.2_

PR #72, opened 2022-07-11
bugfix: allow false value on required bool fields

krushiraj:


ganigeorgiev:
Hm, maybe I should document this, but the current behavior is intentional. The same is true for text fields when the input is `""`.

Consider for example having a collection with "I agree checkbox" field. It's a lot easier for the developer to just check the field as required and expect only `true` to pass create/update validations.

ganigeorgiev:
I don't understand. If the boolean field was marked as required you need to check it. That should be the current behavior. If you want a boolean field to support both `true` or `false` values, then just don't mark it as required.

Boolean fields can have only `true` or `false` value. Even if we allow a validator exclusion,  we always cast the value to bool - https://github.com/pocketbase/pocketbase/blob/master/models/schema/schema_field.go#L238.

ganigeorgiev:
@krushiraj On second thought you have a point. I may need to revisit the `PrepareValue` caster because currently the other fields allow `null` (and yes, you can not set in the admin a text field to null, so that will need to change or we'll have to remove the null compatibility). I'll have to think a little more on this.

ganigeorgiev:
@krushiraj you may want to check https://github.com/pocketbase/pocketbase/issues/122

PR #81, opened 2022-07-11
Update README

TLC-10:
I am currently practicing pull requests.

Updated your bullet points to follow styling.
Corrected the support email.


ganigeorgiev:
Thank you but there is no need for this change. Let's keep the bullets lowercase (eventually later I'll remove the `and ...` prefix from the last bullet). The support email is written like this to prevent bots indexing it.

PR #82, opened 2022-07-12
Simplified version flag.

ValleyZw:
- Used cobra `Version` for simplicity: 'A shorthand "v" flag will also be added if the command does not define one.'
- Removed original version flag.


ganigeorgiev:
There were 2 reasons why I din't use the `Version` of the command:
- I wanted it to have the same API as for example `go run version`
- the `--version` flag is not passed by default to the child commands (which is inconsistent with the other global flags)`

But on second thought, I think as it is in the PR maybe is better because its a lot more simple and this way users will have the option to change the printed name of the application more easily since they'd have access to PocketBase.RootCmd.Use (additionally we may want to remove the `(untracked text)` and leave it empty by default, I'll think a little more on this later).

Issue #37, opened 2022-07-09
Internationalization Support


Tim-Paik:
I didn't find a way to switch languages on the official website, documentation, dashboard, etc. Maybe one would be nice? Great projects need international support!

gedw99:
I use i18n for golang backend. Dates and Tines and currency need to be handled too. 

Have not checked if db late stores datetimes in Zulu tine and convers to uses local for gets and the opposite for sets to the db . Anyone know if it does ?

gedw99:
The standard golang way is also good . Explained ere: https://www.alexedwards.net/blog/i18n-managing-translations

ganigeorgiev:
Internationalization for the error responses could be done client side because each error comes with a fixed key `code` that could be used as a translation key. Internationalization for the user defined data is left to the users, either by creating a multiple fields (eg. `enTitle`, `deTitle`, etc.) or by maintaining each language data in its own collection.

About the internationalization for the public resources - coming from maintaining [Presentator](https://github.com/presentator/presentator), which is pretty much stable and where we support more than 8 languages via [Transifex](https://www.transifex.com/presentatorio/web-platflorm/), it's really hard to maintain and verify users' contributions (for documentation will be a nightmare to support).

PocketBase is still in its very early development stage and a lot of things would change making the translations irreverent.

I'll close the issue for now, but when we reach v1.0.0 will definitely reconsider it and will start a discussion with the community to see how many people are willing to help.

iskaa02:
As you proposed having something `enTitle`,`deTitle` would work just fine for bilingual application but application that support over 4 languages it will be hard to maintain and use.

what i thought about is having a new field type called 'i18n' and it will be JSON, each key will represent a language.
```json
{
  "en":"support for more than one language"
  "ar":"دعم لأكثر من لغة"
}
```
 when resolving records the language will be taken from Http header and there should be a fallback language when the language on the header not supported.
this would make it more convenient and easier to use.

Not sure if you think if this is good fit for the project, or specialized use case, I've already built a very basic version that fits my needs, if you want we can work on this together.

ganigeorgiev:
[...]

PR #85, opened 2022-07-12
Add i18n to the Web UI

2g8:
1, using svelte-i18n
2, lang files in public/lang/*.json
3, Something in record and collection need continue work

ps: sorry for the ignore and lock files, don't know how to reverse them


ganigeorgiev:
Hi thanks for the PR. I see that you put a lot of effort in it, but I mentioned in issue #37,  at the moment I don't want to add internationalization support because it is hard to maintain/verify and the admin UI is just not stable yet - a lot of things could change, new things will be added making the translations irrelevant/outdated.

lifetin:
It's been three years, but there's still no support

PR #88, opened 2022-07-12
Change timing of record init for RecordUpsertPanel

Atozzio:
Thanks for the great project!
When adding new records to a collection, if the preset fields contain a 'file' field. The user will get a reminder saying that the user may have unsaved changes even though nothing has changed.
image

I changed the timing of the record initialization and this might help with it.
Do you mind taking a look at this pr at your convenience? Thank you so much in advance!

### DIFF
--- a/ui/src/components/base/OverlayPanel.svelte
+++ b/ui/src/components/base/OverlayPanel.svelte
@@ -86,6 +86,8 @@
             oldFocusedElem = document.activeElement;
             wrapper?.focus();
             dispatch("show");
+            await tick();
+            dispatch("updateInitRecord");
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
Hi thanks for the PR.

Your changes may work but it is not very reliable this way (in addition, we already are triggering `show`, there shouldn't be a need to add a special record specific event in the base overlay component which is used by all popup/overlay panels in the app).

I'll have a look at this later this week, but the reason this happening is because when we load the record model in the fields, some of them actually mutate their binded value by setting default (for example `null` -> `''`, etc.).

PR #89, opened 2022-07-12
simplified some code by returning early and reducing local variable scopes.

ValleyZw:
- Simplified code through early return.
- Replaced the variable name `strings` (conflicted with **strings** package) with `result`.
- Reduced some local variable scopes.
- Changed JWT initialization method. The previous method will be deprecated: 'Deprecated: In future releases, this field will not be exported anymore and should be set with an option to `NewParser` instead.'


ganigeorgiev:
Thanks for the cleanup!

PR #91, opened 2022-07-12
Video file upload and preview

thomasdseao:
- Add video types on collection creation.
- Add video preview popup as existing for image.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1203,12 +1203,7 @@
       "resolved": "https://registry.npmjs.org/@codemirror/autocomplete/-/autocomplete-6.0.4.tgz",
       "integrity": "sha512-uP7UodCRykPNwSAN+wYa/AS9gJI/V47echCAXUYgCgBXy3l19nwO7W/d29COtG/dfAsjBOhMDeh3Ms8Y5VZbrA==",
       "dev": true,
-      "requires": {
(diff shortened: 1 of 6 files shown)


ganigeorgiev:
Thanks for the PR. I'll reviewed it this weekend and will let you know if something need to change (I'm only concerned about the loading performance, but will have to tested).

ganigeorgiev:
@thomasdseao I've just tested it locally, but I confirmed my concerns about the loading performance. It seems that with the current setup we are downloading the video (or some part of it; on chrome the default is 25sec) just to show the preview thumb.

There is the `preload="metadata"` attribute that we can use, but it's behavior is not unified (see https://web.dev/video-and-source-tags/#preload) on all platforms and you still need to download some portion of the video, which may not be OK if you are listing hundreds of records with video files (even smaller 10sec videos may require more than 1MB to download each).

If we want to show a video thumb, we'll have to generate it as image and cache it server side, but for now, although nice to have, I'd consider this is a very low priority feature, because I'm not sure how many users are actually uploading and serving videos directly from their servers (I have past experience with developing video streaming apis for vod and live content and usually this is done with hls, dash or something else more efficient than just directly serving the video source file in the browser).

I'll close the PR and will add only the MimeType video presets to the file field options.

PR #110, opened 2022-07-14
Strava as OAuth provider + UI Example

mbecker:
Hi,

my use case is to use Strava as an OAuth-Provider for authentication. In general, quite easy to implement a new OAuth-Provider. Additionally, I've updated the Admin UI (no Strava logo) and adding an example "ui-auth" to test the oauth implementation.

**But the implementations includes one hack:**

> The Strava OAuth-Userinfo-Endpoint does not returny any email address. The current sign up / sign in implementations requires a valid email address.
> The implementation creates a fake email address as follows: STRAVAID@strava.com
> This means, the OAuth-user can not natch to any existing user and any email sent to the user will not work

Additional feature request for later implementation:

**General OAuth-Provider**
- Implement a general OAuth-Provider without any names, URLs and scopes
- The general OAuth-Provider may be initiated multiple times
- Let the user initiate the general OAuth-Provider multiples times; the identifier should be the name given in the setting by initiating the provider
- Let the user decide how to specify the scopes (see comma separated list of scopes)
- Let the user decide how to extract the user's information

**Registering via OAuth without any email address**
- The current implementation for authentication (registration / signin) is based on an unique identifier which is the email address
- As seen at the Strava implementation some OAuth-providers do not return any email address by design; this results in an error at the OAuth flow (registration fails)

Because the email is used as the primary unique identifier at the signin-process I would propose something as follows:

> Let the admin define a setting "sign up / sign in without email address" 
> -> The OAuth-user is allowed to sign up / sign in without any email address"

>Let the admin define a setting "oauth users without any email address must verify their email"
 > -> The OAuth-user signs up without any email address -> The user is created but marked as "email not verified/blank" -> 
 The sign in API returns: User must verify their email address (with an hash key to enter the email address later)
 
I would let the user sign in via the OAuth-provider and it's the user's / ui's responsibility to add an email address. An example may be Keycloak which let's you to configure that the user needs to enter an email address.


ganigeorgiev:
I'm sorry but I'll reject this PR.
Similar to #111 this is too hacky to be part of the main codebase and it's very specific to your use case.

If a OAuth2 provider doesn't return an email address then this is not supported at the moment. Each user need to have a valid email address, because:
- if the user decide to change his/her email, then the hack in the PR will not work
- we need to be able to link the user account when different providers are used at the same time
- users need to have an option to reset their password somehow

I'm not planning to support different user identifiers (username, ids, etc.) at the moment. This is left to handle in user land code.

PR #111, opened 2022-07-14
OAuth 2.0 Login with access token

mbecker:
Hi,

I'm using NextJS and [NextAuth.js](https://next-auth.js.org/) with an OAuth provider (Strava). As document at NextAuth.js the library includes some callbacks for using an OAuth-provider to hook into the login process. However, it does not include a callback to get the code returned by the provider to request the access token.

This pull request implements a pocktebase-oauth-login with an access token sent by the user rather requesting the access token with a code by itself.

To have not a major change for the JS library I've used some convention:
> The redirectURL in the JS method "client.Users.authViaOAuth2" must include the oauth2.Token struct as query params as follows: http://localhost:3000?access_token=XXX123&refresh_token=ABC123&token_type=Bearer&expires_at=123344

- Only the query param "access_token" is a must have
- The host is not relevant for but must be valid address to be validated by the original implementation
- (The other token details are used for later implementation to store the token in the user's profile)

The JS is as follows:
```javascript
client.Users.authViaOAuth2('strava', `NOT_RELEVANT`, 'NOT_RELEVANT', `http://example.com?access_token=${account.access_token}&refresh_token=${account.refresh_token}&token_type=${account.token_type}&expires_at=${account.expires_at}`)
        .then((userData) => {
          console.log("Pocketbase user data: ", JSON.stringify(userData, undefined, 2))
          return true;
        })
        .catch((err) => {
          console.log('Error requesting auth / signin at pocketbase: ', err);
          return false;
        });
```

Feature request for better implementation:
- Implement an own API endpoint which takes the oauth2.Token as the input for querying the user's detail
- Implement the new endpoint in the JS library

Unfortunately, this pull request includes the other one as well. I've created this branch based on the "oauth strava"-branch; maybe I should create a branch based in the main-branch?


mbecker:
Hi,

I think "hacky" is the right word for the implementation :-) Thanks for your feedback and I do understand your feedback.

Besides that, I've managed to use NextAuth.js to use it with PocketBase the correct way.

But, may I ask: With NextAuth.js I can do the PKCE flow on the client side; isn't that safe?

Great work and thanks, again!

ganigeorgiev:
It is safe to make request client side to a OAuth2 PKCE service because the client side doesn't hold the OAUTH2 access token.
Here is a great article with example SPA integration using the PKCE flow - https://www.taniarascia.com/oauth-pkce-authorization/

PR #118, opened 2022-07-14
Twitter Oauth2 Provider

thomasdseao:
Add Twitter Oauth2 provider.
**The user should append the redirect_uri with that one specified on the twitter app because Twitter check it and reject the request if it is not matching.**


tools/auth/twitter.go
```diff
@@ -0,0 +1,49 @@
...
+		scopes:     []string{"users.read"},
+		authUrl:    "https://twitter.com/i/oauth2/authorize",
+		tokenUrl:   "https://api.twitter.com/2/oauth2/token",
+		userApiUrl: "https://api.twitter.com/1.1/account/verify_credentials.json?include_email=true",
+	}}
+}
+
+// FetchAuthUser returns an AuthUser instance based on the Twitter's user api.
+func (p *Twitter) FetchAuthUser(token *oauth2.Token) (*AuthUser, error) {
+	// https://developer.twitter.com/en/docs/twitter-api/users/lookup/api-reference/get-users-me
```
ganigeorgiev, line 29: Could you  replace the url with the one that has the acount fields - https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials#example-response ?

tools/auth/twitter.go
```diff
@@ -0,0 +1,49 @@
...
+	rawData := struct {
+		Id                   string
+		Name                 string
+		Email                string
+		ProfileImageUrlHttps string
```
ganigeorgiev, line 34: `ProfileImageUrlHttps` will not be unmarshalized because in the [twitter response](https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials#example-response) it is returned as `profile_image_url_https`.

Please add json tags to the struct members mapped to the response fields (id, name and email including for consistency).

ganigeorgiev (changes requested):

ganigeorgiev:
Hi, sorry for the delay.

I've just pull-ed changes locally and tried to test the integration, but without success.
There are several things that came up:
1. we need to update `BaseProvider.FetchRawUserData` to return an error when the client fails with >=400 response, since it turned out `http.Client.Get()` returns an error only when somethings wrong with the connection (aka. non 2xx responses will be not returned as `err`), so we need to manually check for >= 400 status codes and return the error to the user in order to improve error reporting

2. After enabling the above error log, I found out that you'll need "Elevated Access" in order to request user api url:
```json
{
    "errors":[{"message":"You currently have Essential access which includes access to Twitter API v2 endpoints only. If you need access to this endpoint, you’ll need to apply for Elevated access via the Developer Portal. You can learn more here: https://developer.twitter.com/en/docs/twitter-api/getting-started/about-twitter-api#v2-access-leve","code":453}]
}
```
3. After enabling it, and tried again, I got the following error:
```
FetchRawUserData for https://api.twitter.com/1.1/account/verify_credentials.json?include_email=true (403): {
  "title": "Forbidden",
  "type": "about:blank",
  "status": 403,
  "detail": "You are not permitted to use OAuth2 on this endpoint"
}
```
So unless the message is not misleading, it seems that we cannot use `https://api.twitter.com/1.1/account/verify_credentials.json?include_email=true` to get the user email address.

Maybe I've messed up something while creating the Twitter OAuth2 app (I also enabled the OAuth1 option in order to check the "Request email from users" checkbox):
image info

**@thomasdseao Were you able to test it from start to end?**

thomasdseao:
Hi @ganigeorgiev,

Yes that's a known problem because the verify_credentials don't have equivalent with oauth2 : https://twittercommunity.com/t/which-api-to-get-user-after-oauth2-authorization/162417/33

At the moment it is not possible to retrieve the user data with oauth2 until they develop and deploy this feature.
So we have to wait the compatibility of this endpoint.

Let's keep this PR open until they add the compatibility (we will just have to change the URL).

thomasdseao:
For the error handling, how do you want to proceed if providers return errors in differents fields ? 
That mean we can not retrieve the error message by the same way for every providers.

ganigeorgiev:
@thomasdseao Thanks for the follow up.

I'll close the PR since I'm planning changes to the internals and will be hard to keep it in sync (currently the tests also doesn't pass because the test database need to be updated with the twitter settings).

In addition to the endpoint, we'll have to change probably the unmarshallized fields too if they decide to use different response.

I'll apply the BaseProvider error reporting in different commit. There is no need to statically type the errors, we just need to return an error so that it could be logged in the console when `--debug` is used.

thomasdseao:
I understand. One question about the design, do you use some UI kit ?

ganigeorgiev:
No, the UI is extremely simple and I didn't want to bring some heavy and full featured library (the styles could be found here - https://github.com/pocketbase/pocketbase/tree/master/ui/src/scss).

The styles are not tightly coupled with the svelte components, because initially I had the idea to make the interface pluggable with external user-land code, so that users can customize it without requiring a build step, but for now I decided to keep it simple for the first version and leave it for later to consider depending on the user feedback.

ganigeorgiev:
@thomasdseao With [v0.6.0 release] PocketBase now supports OAuth2 clients that don't return an email address so I've reintegrated the Twitter OAuth2 provider.

PR #126, opened 2022-07-15
simplified UcFirst and Sentenize functions and removed redundant character escapes in regex.

ValleyZw:
- simplified `UcFirst` and `Sentenize` functions.
- removed redundant character escapes in regex.
- added more tests for `UcFirst` and `Sentenize`.


tools/inflector/inflector.go
```diff
@@ -6,18 +6,17 @@ import (
...
+var (
+	columnifyRemoveRegex = regexp.MustCompile(`[^\w.*\-_@#]+`)
+	snakecaseSplitRegex  = regexp.MustCompile(`[\W_]+`)
+)
 
 // UcFirst converts the first character of a string into uppercase.
 func UcFirst(str string) string {
-	if str == "" {
-		return ""
+	for i, c := range str {
```
ganigeorgiev, line 16: This is my opinion makes it harder to read and a little confusing at first. Let's keep the old implementation.

tools/inflector/inflector.go
```diff
@@ -31,16 +30,14 @@ func Sentenize(str string) string {
 	if str == "" {
 		return ""
 	}
+	str = UcFirst(str)
```
ganigeorgiev, line 33: (nitpick) please add a newline before the assignment to make it more clear visual that is after the if and return block.

tools/inflector/inflector.go
```diff
@@ -6,18 +6,17 @@ import (
 	"unicode"
 )
 
-var columnifyRemoveRegex = regexp.MustCompile(`[^\w\.\*\-\_\@\#]+`)
-var snakecaseSplitRegex = regexp.MustCompile(`[\W_]+`)
+var (
+	columnifyRemoveRegex = regexp.MustCompile(`[^\w.*\-_@#]+`)
```
ganigeorgiev, line 10: This doesn't look like the same pattern. For example, `.` and `*` should remain unescaped because we want to preserve strings like `field.test` or `field[*]` (this is for a future feature for querying in json arrays).

Let's keep the original. If your change passes the tests, then probably we need to improve them.

ganigeorgiev (changes requested):

Issue #122, opened 2022-07-15
Nullable and non-nullable filelds


ganigeorgiev:
Currently the null handling behavior is not very clear - with the exception of bool and multiple value fields, all other fields support storing and returning `null`, if the field is not marked as "Required".

We could improve and normalize this behavior by introducing a new **"Nullable"** boolean field option, allowing developers to specify whether they want `null` or zero-values in their api responses. In general, we want to support the following field return values:

| Field               | Nullable response values                         | Non-nullable response values                   |
|:--------------------|:-------------------------------------------------|:-----------------------------------------------|
| Text                | `null`, `""`, `"test"`                           | `""`, `"test"`                                 |
| Number              | `null`, `0`, `-1`, `1`,`1.5`                     | `0`, `-1`, `1`, `1.5`                          |
| Bool                | `null`, `false`, `true`                          | `false`, `true`                                |
| Email               | `null`, `""`, `"test@example.com"`               | `""`, `"test@example.com"`                     |
| Url                 | `null`, `""`, `"https://example.com"`            | `""`, `"https://example.com"`                  |
| Date                | `null`, `""`, `"2022-01-01 00:00:00.000"`        | `""`, `"2022-01-01 00:00:00.000"`              |
| Select (single)     | `null`, `""`, `"optionA"`                        | `""`, `"optionA"`                              |
| Select (multiple)   | `null`, `["optionA", "optionB"]`                 | `[]`, `["optionA", "optionB"]`                 |
| File (single)       | `null`, `"12i...ZjL.png"`                        | `""`, `"12i...ZjL.png"`                        |
| File (multiple)     | `null`, `["12i...ZjL.png", "34i...ZjL.txt"]`     | `[]`, `["12i...ZjL.png", "34i...ZjL.txt"]`     |
| Relation (single)   | `null`, `"74kmig0HEyR7uuD"`                      | `""`, `"74kmig0HEyR7uuD"`                      |
| Relation (multiple) | `null`, `["74kmig0HEyR7uuD", "FtHAW9feB5rze7D"]` | `[]`, `["74kmig0HEyR7uuD", "FtHAW9feB5rze7D"]` |
| User (single)       | `null`, `"74kmig0HEyR7uuD"`                      | `""`, `"74kmig0HEyR7uuD"`                      |
[...]

PR #127, opened 2022-07-15
fix field name input issue

clysto:
When there is already a `file` field in schama, adding a filename field from the admin panel will be automatically recognized as a duplicate field by the `normalizeFieldName` function, so `filename` will become `file1name`.

https://github.com/pocketbase/pocketbase/blob/1095637bcd9aaf3901bdc075dc430a1003efaa8d/ui/src/components/collections/FieldAccordion.svelte#L91-L100

### DIFF
--- a/ui/src/components/collections/FieldAccordion.svelte
+++ b/ui/src/components/collections/FieldAccordion.svelte
@@ -221,9 +221,13 @@
                         spellcheck="false"
                         autofocus={!field.id}
                         value={field.name}
-                        on:input={(e) => {
+                        on:blur={(e) => {
                             field.name = normalizeFieldName(e.target.value);
                             e.target.value = field.name;
+                        }}
+                        on:input={(e) => {
+                            const name = CommonHelper.slugify(e.target.value);
+                            e.target.value = name; 
                         }}
                     />
                 </Field>


ganigeorgiev:
Could you elaborate more on the issue? Name normalization usually should happen only when the user is typing/paste-ing the field name. There shouldn't be a need for the `blur` listener.

Currently, if you have an existing field `"title"`, the normalizer will prevent any new fields to be named the same and will try to auto add a suffix counter to the field name, eg. `"title1"`.

ganigeorgiev:
Ah, I understand what you mean. I've just reproduced it.
But I think the on:blur is not the best solution for this, because the user will see the correct change (if needed) only on blur and not during typing.

clysto:
yeah. I think it would be better to use text to inform the user of the existence of duplicate fields, but that might require adding a help block to each schema field.

ganigeorgiev:
@clysto I'll close the PR for now. I'll think more on it while working on the admin UI changes required for https://github.com/pocketbase/pocketbase/issues/122.
We always have server-side validations as fallback and maybe your suggestion for showing an error for duplicated field(s), instead of automatically suffixing them, may turn out to be a better approach, but I'll have to experiment with it.

ganigeorgiev:
@clysto A slightly better and different duplicated/reserved names handling was added in [v0.3.0 release](https://github.com/pocketbase/pocketbase/releases/tag/v0.3.0).

clysto:
wow, the new admin panel is a lot easier to use, thanks for developing it.❤️

Issue #92, opened 2022-07-12
Add official dockerhub image


EriksonGM:
Just create a official dockerhub account and publish this amazing project.

```
FROM alpine:3.6
ARG POCKETBASE_VERSION=0.1.2

# Install the dependencies
RUN apk add --no-cache \
    ca-certificates \
    unzip \
    wget \
    zip \
    zlib-dev

# Download Pocketbase and install it for AMD64
ADD https://github.com/pocketbase/pocketbase/releases/download/v${POCKETBASE_VERSION}/pocketbase_${POCKETBASE_VERSION}_linux_amd64.zip /tmp/pocketbase.zip
RUN unzip /tmp/pocketbase.zip -d /usr/local/bin/
RUN chmod +x /usr/local/bin/pocketbase

# Notify Docker that the container wants to expose a port.
EXPOSE 8090

# Start Pocketbase
CMD [ "/usr/local/bin/pocketbase", "serve" ]
```

This is an example made for [bscott](https://github.com/bscott) / [pocketbase-docker](https://github.com/bscott/pocketbase-docker) credits to him.

KROSF:
here one with multiarch support https://github.com/muchobien/pocketbase-docker

bscott:
I'll update mine for Multi-Arch - Be aware of #41

gaby:
@EriksonGM @bscott Can you uodate the base image to alpine 3.15? The version you are using is several years old.

Latest release notes are here: https://www.alpinelinux.org/posts/Alpine-3.13.12-3.14.8-3.15.6-3.16.2-released.html

insuusvenerati:
Hey all! Despite my best efforts, I haven't been able to get pocketbase to work in Docker. With a very basic Dockerfile I always only get `curl: (56) Recv failure: Connection reset by peer` or `ERR_CONNECTION_RESET` in the browser. 

```
FROM alpine:3.16

ADD https://github.com/pocketbase/pocketbase/releases/download/v0.7.2/pocketbase_0.7.2_linux_amd64.zip /tmp/pocketbase.zip
RUN unzip /tmp/pocketbase.zip -d /usr/local/bin \
    && chmod +x /usr/local/bin/pocketbase \
    && rm /tmp/pocketbase.zip

RUN adduser -D pocketbase pocketbase
USER pocketbase
WORKDIR /home/pocketbase

EXPOSE 8090

CMD ["pocketbase", "serve"]
```

ganigeorgiev:
[...]

PR #135, opened 2022-07-15
creates multiarch compatible Dockerfile

kennethklee:
this would be a step towards #92 .

tested with: `docker buildx build --platform linux/arm/v7,linux/arm64/v8,linux/amd64 -t pocketbase .`

i'm not familar with darwin builds, but you should be able to extend the above to it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+# Build
+# =====
+FROM golang:1.18.4-alpine AS gobuilder
+
+WORKDIR /app
+
+COPY . .
+
+# install git for version
+RUN apk add --no-cache git
+
+RUN cd examples/base \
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Hi, thanks for the PR but I don't think this is needed.

The image in the PR seems to be more suitable when you want to build the distributed binaries for different platforms, but we don't need that because we already have gorealeaser taking care for it - https://github.com/pocketbase/pocketbase/blob/master/.goreleaser.yaml

My idea for the production docker image was a lot more simpler - use alpine or some other small linux distro and just download the linux archive with the tagged image version -> extract the zip -> start the binary exposing the default port.

In any case, let's not rush this. The implementation is easy, maintaining it is hard (especially when Docker Hub no longer provide *free automated builds).

_* They have open source offerings, but from my experience maintaining Presentator (see [#178](https://github.com/presentator/presentator/issues/178#issuecomment-890749782)), the communication with the Docker Hub intern, that was reviwing my case at the time was very difficult and after 3 weeks of back-and-forth mails exchange, I ended up building the tags locally._

PR #149, opened 2022-07-17
cleaned up some code, reused UcFirst function and fixed doc and code inconsistency.

ValleyZw:
- cleaned up some code.
- added slice initial caps.
- reused `UcFirst` function.
- added `UcFirst` and `Sentenize` test cases.
- fixed doc and code inconsistency.


apis/middlewares.go
```diff
@@ -160,23 +160,25 @@ func LoadAuthContext(app core.App) echo.MiddlewareFunc {
...
+			}
+
+			rolePrefix = "Admin "
+			if strings.HasPrefix(token, rolePrefix) {
+				admin, err := app.Dao().FindAdminByToken(
+					token[len(rolePrefix):],
+					app.Settings().AdminAuthToken.Secret,
+				)
+				if err == nil && admin != nil {
+					c.Set(ContextAdminKey, admin)
```
ganigeorgiev, line 181: This change may not be necessary, since I don't think there is a need to check the token prefix if is empty and 2 times in that. Let's keep the original (eventually we may want to return early to reduce the nesting, but the function is small and it doesn't hurt the readability too much)

tools/filesystem/filesystem.go
```diff
@@ -113,23 +113,20 @@ func (s *System) Delete(fileKey string) error {
...
-		return failed
+		return []error{errors.New("Prefix mustn't be empty.")}
 	}
 
-	dirsMap := map[string]struct{}{}
-	dirsMap[prefix] = struct{}{}
-
-	opts := blob.ListOptions{
-		Prefix: prefix,
+	failed := make([]error, 0)
```
ganigeorgiev, line 120: Since we don't have a cap, let's use the literal (eg. `failed := []error{}`) in this case for consistency with the dirsMap.

tools/filesystem/filesystem.go
```diff
@@ -143,17 +140,17 @@ func (s *System) DeletePrefix(prefix string) []error {
 
 		if err := s.Delete(obj.Key); err != nil {
 			failed = append(failed, err)
-		} else {
-			dirsMap[filepath.Dir(obj.Key)] = struct{}{}
+			continue
 		}
+		dirsMap[filepath.Dir(obj.Key)] = struct{}{}
```
ganigeorgiev, line 145: Let's add a new line after the if block for easier visual distinction.

daos/record.go
```diff
@@ -154,26 +154,25 @@ func (dao *Dao) FindUserRelatedRecords(user *models.User) ([]*models.Record, err
 		return nil, err
 	}
 
-	result := []*models.Record{}
+	result := make([]*models.Record, 0, len(collections))
```
ganigeorgiev, line 157: This may not be necessary, since the cap most of the cases will be always wrong and will have to expand anyway because we are filling the result with the found records. Let's keep the literal version (in my opinion is easier to read too).

ganigeorgiev (changes requested):
Thanks for the PR!

_I left some nitpicking comments you could check when you find time._

PR #156, opened 2022-07-18
Support S3 forcePathStyle for systems like minio

mbessette94:
- Added .devcontainer for vscode development in docker
- Added support for `forcePathStyle` on S3 Driver
- Tested with minio uploading Avatars and Files onto collection records

Let me know if you'd prefer the toggle moved
Screenshot from 2022-07-17 21-46-03


core/settings.go
```diff
@@ -260,6 +261,7 @@ func (c S3Config) Validate() error {
 		validation.Field(&c.Region, validation.When(c.Enabled, validation.Required)),
 		validation.Field(&c.AccessKey, validation.When(c.Enabled, validation.Required)),
 		validation.Field(&c.Secret, validation.When(c.Enabled, validation.Required)),
+		validation.Field(&c.ForcePathStyle, validation.When(c.Enabled, validation.Required)),
```
ganigeorgiev, line 264: This line could be removed. The `ForcePathStyle` should be optional, not required.

ui/src/components/settings/PageStorage.svelte
```diff
@@ -130,6 +130,12 @@
                                 <input type="text" id={uniqueId} required bind:value={s3.endpoint} />
                             </Field>
                         </div>
+                        <div class="col-lg-12">
+                            <Field class="form-field form-field-toggle" name="s3.forcePathStyle" let:uniqueId>
+                                <input type="checkbox" id={uniqueId} required bind:checked={s3.forcePathStyle} />
+                                <label for={uniqueId}>Force Path Style?</label>
+                            </Field>
+                        </div>
```
ganigeorgiev, line 138: This checkbox should be optional and not required.
Let's also move it below under all the other S3 fields and make it a regular checkbox (aka. remove the `form-field-toggle` class).
Let's also remove the question mark from the label and add a info icon with tooltip explaining what this option do.
Or in other words, place the below code, before `<!-- margin helper -->`
```html
// add the tooltip action in the imports section
import tooltip from "@/actions/tooltip";

....

<div class="col-lg-12">
    <Field class="form-field" name="s3.forcePathStyle" let:uniqueId>
        <input type="checkbox" id={uniqueId} bind:checked={s3.forcePathStyle} />
        <label for={uniqueId}>
            <span class="txt">Force path-style addressing</span>
            <i
                class="ri-information-line link-hint"
                use:tooltip={{
                    text: 'Enable to force the request to use path-style addressing, eg. "https://s3.amazonaws.com/BUCKET/KEY" instead of the default "https://BUCKET.s3.amazonaws.com/KEY".',
                    position: "top",
                }}
            />
        </label>
    </Field>
</div>
```

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
+
+# [Optional] Uncomment the next line to use go get to install anything else you need
+# RUN go get -x <your-dependency-or-tool>
+
+# [Optional] Uncomment this line to install global node packages.
+# RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
```
ganigeorgiev, line 18: Let's not add this since it is too specific to the user environment and for vscode only (there also too many optional things that I haven't tested).

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,58 @@
...
+	// "otherPortsAttributes": {
+	// 		"onAutoForward": "silent"
+	// },
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "go version",
+
+	// Uncomment to connect as a non-root user. More info: https://aka.ms/vscode-remote/containers/non-root.
+	"remoteUser": "vscode"
+}
```
ganigeorgiev, line 58: Same as above.

We may add it in the future, but let's keep it out of this PR.

ganigeorgiev (changes requested):
Hi, thanks for the PR!

I left some notes, you could review them when you have time.

I'm usually against this option, because it is [deprecated by amazon](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) and probably they will remove it from the library too in future versions, but OK. Let's keep it for better compatibility with other S3 providers.

ganigeorgiev:
In addition, also make sure to run/update the Go tests - see https://github.com/pocketbase/pocketbase/blob/master/CONTRIBUTING.md#making-changes-in-the-go-code

ui/src/components/settings/PageStorage.svelte
```diff
@@ -130,6 +130,12 @@
                                 <input type="text" id={uniqueId} required bind:value={s3.endpoint} />
                             </Field>
                         </div>
+                        <div class="col-lg-12">
+                            <Field class="form-field form-field-toggle" name="s3.forcePathStyle" let:uniqueId>
+                                <input type="checkbox" id={uniqueId} required bind:checked={s3.forcePathStyle} />
+                                <label for={uniqueId}>Force Path Style?</label>
```
Lex-2008, line 136: Maybe also remove question mark here?

Note that I'm not a maintainer here, so up to @ganigeorgiev to decide :-)

But I think that usually options don't have question marks - see, for example, "Use S3 storage" option on your screenshot.

Lex-2008 (commented):
nice! Without it I had to use `MINIO_DOMAIN` env variable and add bucket name to DNS (or /etc/hosts), see #90. And with this option it will just work!

ui/src/components/settings/PageStorage.svelte
Matt-Bessette: Updated per @ganigeorgiev's phrasing of it :D I'm not much of a ux person, I'm used to being told where to put stuff!

ganigeorgiev:
Thank you!

I've merged it locally since the tests had to be updated. It will be included in the next minor bugfix/feature release.

PR #163, opened 2022-07-18
Fix for all migrations reverted when doing `migrate down 1`

### LINKED ISSUE #161: migrate down always reverts all migrations
While testing a migration, noticed `pocketbase migrate down 1` reverts all migrations including the init one.

From a peek in the code, `totalReverted` never gets incremented. perhaps could change it to `len(applied)`?

https://github.com/pocketbase/pocketbase/blob/f56adf26f484bfa45f9bd019ccb0195790e186ab/tools/migrate/runner.go#L200

kennethklee:
This should fix #161 

Debated using `len(applied)` in the if statement, but didn't want to remove `totalReverted` since I'm not too familiar.

### DIFF
--- a/tools/migrate/runner.go
+++ b/tools/migrate/runner.go
@@ -220,6 +220,7 @@ func (r *Runner) Down(toRevertCount int) ([]string, error) {
 				return fmt.Errorf("Failed to save reverted migration info for %s: %w", m.file, err)
 			}
 
+			totalReverted++
 			applied = append(applied, m.file)
 		}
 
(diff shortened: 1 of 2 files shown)


tools/migrate/runner.go
```diff
@@ -220,6 +220,7 @@ func (r *Runner) Down(toRevertCount int) ([]string, error) {
 				return fmt.Errorf("Failed to save reverted migration info for %s: %w", m.file, err)
 			}
 
+			totalReverted++
```
ganigeorgiev, line 223: Let's remove this counter. It'll be better to rename the `applied` variable and change it declaration to `reverted := make([]string, 0, toRevertCount)` and use the `len(reverted)` instead.

ganigeorgiev (changes requested):

PR #204, opened 2022-07-24
fixes multiple ? in url and simplifies query string logic

kennethklee:
Found an issue when there's multiple `?` in the URL, the query string starts to repeatedly append.

for example: `http://localhost:8090/_/?#/collections?sort=-created&collectionId=106kVGvE2uxfz9X?sort=-created?collectionId=106kVGvE2uxfz9X&sort=-created?collectionId=106kVGvE2uxfz9X&sort=-id` and can keep building.

Functionality wise, it still works fine. Not sure how I go to `http://localhost:8090/_/?` in the first place but somehow I got there.

While debugging, noticed that arrays don't translate well so fixed that. And simplified the logic with native UrlSearchParams instead of string building.

Here's a note:
It looks like pocketbase builds it's URL with the querystring after the hash (since pocketbase uses hashbang for routing, it works, but isn't typical URL structure). The `getQueryString` function looks to get the query string after the hash. It can also get the query string before, but will get to an Error state if that happens (endless loop). So I removed that functionality. To add it back, you'd change line 669 to `return location.search || location.hash.split('?')[1] || '';`, but that will continue to break pocketbase.

Another tidbit:
With this, there is no harm in having a `?` prefix the query string anymore in `getQueryString`, because UrlSearchParams will handle that.

### DIFF
--- a/ui/src/utils/CommonHelper.js
+++ b/ui/src/utils/CommonHelper.js
@@ -665,15 +665,8 @@ export default class CommonHelper {
      * @return {String}
      */
     static getQueryString(url) {
-        let queryStartPos = url.indexOf("?");
-
-        if (queryStartPos < 0) {
-            return "";
-        }
-
-        let hashStartPos = url.indexOf("#");
-
-        return url.substring(queryStartPos + 1, (hashStartPos > queryStartPos ? hashStartPos : url.length));
+        let location = url ? new URL(url) : window.location;
+        return location.hash.split('?')[1] || '';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Hm, this is actually a little more complex than it seems. It is a result of a very early version of the admin UI, where initially I was using pretty urls and the later implementation wasn't adapted to the current router.

The position of the query string actually matters because if we put it before the hash, then every time when we call `window.location.replace` it will reload the page, so if we end up keeping it before the hash we also will have to update that portion of the code and to use the history api (ideally the position of the query string shouldn't affect the behavior of the application).

This will need to be fixed differently. After the Dart SDK, I'll try to squeeze some time and will rework the query routing (it may turn out that this part of the code may not be even necessary anymore).

kennethklee:
I can understand that.

While making this PR, I had to fight the urge to simplify the page routing -- which would have made this PR redundant.

Let me know if you want a hand in that.

ganigeorgiev:
@kennethklee Just to let you know that the manual query string parsing was removed [v0.3.4 release](https://github.com/pocketbase/pocketbase/releases/tag/v0.3.4) and we are now using `URLSearchParams` with the readonly svelte-spa-router's querystring store.

kennethklee:
huzzah! awesome

PR #207, opened 2022-07-24
use read-only scopes for github auth

satotake:
According to github's doc, `user` scope includes write access to profile info.
Read-only scope would be sufficient for this auth flow.

ref: https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps


ganigeorgiev:
Thank you. Yes, `"read:user"` and `"user:email"` are sufficient enough. I've just tested with a demo GitHub OAuth2 app, both with public and private email (to trigger the additional `/emails` api call), and it works as expected.

The change will be included in v0.4.0 release.

PR #250, opened 2022-08-01
added armv7 on linux as a build target

yinshanyang:
Added `armv7` for Linux as a build target to support distribution on Raspberry Pi 2 and above.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -20,13 +20,20 @@ builds:
     goarch:
       - amd64
       - arm64
+      - arm
+    goarm:
+      - 7
+    ignore:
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Thanks! It will increase a little the build time, but other users also requested it in the past so I think it is OK in this case.
The target will be available with the next v0.4.0 release.

yinshanyang:
Cheers, looking forward.

ganigeorgiev:
@yinshanyang The arm build target is available with the latest bugfixes [v0.3.4 release](https://github.com/pocketbase/pocketbase/releases/tag/v0.3.4).

Issue #246, opened 2022-08-01
Feature Request: Allow the id field to be provided from the client


robitar:
Firstly, thanks so much for putting this together, I've been playing around with it and it's really quite a thing!

As far as I can see, the id field is automatically generated when a record is created. If an `id` is provided in the request, its ignored. I'm putting together a demo of a local-first/CRDT data structure, which functions offline, and syncs records to the server when it comes back etc.

I can add my own 'key' field which is unique and required, but I still need to do an insert to get a hold of the generated id for any cross references, which is rather awkward. Would it be possible to support client-provided ids without too much churn?

ganigeorgiev:
Allowing clients to set the `id` manually may have some unexpected side-effect when the id is used with some filter operators (see https://github.com/pocketbase/pocketbase/discussions/230#discussioncomment-3232141).

I'll have to think a little more on this, but support for client-provided ids will be added in the near future (this most likely will be also needed for https://github.com/pocketbase/pocketbase/issues/48).

robitar:
Yes makes sense, I hadn't considered the data import scenario.

It also occurred to me that there's currently no 'sequence' value (like an autoincrement id etc) that gives you a stable monotonic row number. Something that lets you say "the last row I saw was x, give me everything greater than x". You can use a combination of creation and modification dates for this, but then you have to rely on the server clock never drifting backwards.

This is also useful for doing cursor style pagination.

Perhaps if a collection had a key generation strategy? Then you could choose between uuid/random, client provided, or autoincrementing sequence?

(Is it worth considering some kind of sequential uuid generation scheme as well? if tables get larger then a random primary can be problematic for performance, I'm not sure if this is true for SQLite, but they are using B trees, so I presume its a factor there)

ganigeorgiev:
@robitar There is actually an internal sqlite sequential key but currently it's not exposed to the API (with the exception for the logs service) - `rowid` (https://www.sqlite.org/rowidtable.html).

robitar:
Ah yes, sounds ideal. But I see what you mean, it would also be tricky to work into the API design, since Id/RowId would be confusing. Re providers, it seems reasonable enough to assume most database providers will support some kind of arbitrary sequence columns outside of primary keys (I'm familiar with MSSQL which finally supports this).

[...]

PR #262, opened 2022-08-03
Set `Id` when save model as ColumnValueMapper

sospartan:
I'm using `ColumnValueMapper` to save model. Accidentally find the map not contains 'id' is still saved, but failed to get it back.
Hope this PR can prevent this scenario.

### DIFF
--- a/daos/base.go
+++ b/daos/base.go
@@ -195,6 +195,7 @@ func (dao *Dao) create(m models.Model) error {
 
 	if v, ok := any(m).(models.ColumnValueMapper); ok {
 		dataMap := v.ColumnValueMap()
+		dataMap["id"] = m.GetId() //prevent Mapper not return 'id'
 
 		_, err := dao.db.Insert(m.TableName(), dataMap).Execute()
 		if err != nil {
--- a/daos/base_test.go
+++ b/daos/base_test.go
@@ -158,6 +158,41 @@ func TestDaoSaveCreate(t *testing.T) {
 	}
 }
 
+type TestAdminMapper struct {
+	models.Admin
+}
+
+func (m *TestAdminMapper) ColumnValueMap() map[string]any {
+	return map[string]any{
+		"email":    m.Email,
+		"avatar":   m.Avatar,
+		"tokenKey": m.TokenKey,
+		"passwordHash":m.PasswordHash,
+	}
+}
+
+func TestDaoSaveCreateWithMapper(t *testing.T) {
+	testApp, _ := tests.NewTestApp()
+	defer testApp.Cleanup()
+
+	model := &models.Admin{}
+	model.Email = "test_new@example.com"
+	model.Avatar = 8
+	mapper := &TestAdminMapper{
+		*model,
+	}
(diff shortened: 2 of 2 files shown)


ganigeorgiev:
Thank you for taking the time preparing the PR, but I don't think it is necessary.

The `ColumnValueMapper` interface was intended for models with dynamic fields and/or when you have different db columns presentation. Or in other words - adding a fixed `id` key in the dao kind of defeat the purpose of the interface. The `id` need to be defined in your model (aka. where the `ColumnValueMap()` method is). You can see how it is done for the `Record` model - https://github.com/pocketbase/pocketbase/blob/master/models/record.go#L208-L218.

I'll close the PR for now, but I'm curious about your use case. What requires you to use it in the first place? If you know the fields of your struct beforehand, you could just use the `db` struct tag, eg:
```go
type MyCustomModel struct {
        BaseModel
 
	Title string `db:"title" json:"title"`
}
```

Also keep in mind that in the near future we will allow the users to provide an `id`  on their own (see https://github.com/pocketbase/pocketbase/issues/246) and that part of the update/create Dao code will change anyway.

ganigeorgiev:
@sospartan Yep, that part of the code will definitely will need to change (I've just noticed that we are using the `id` in the `update()`). I'll have it in mind when reworking the base Dao methods related to the custom insertion id.

sospartan:
> ... but I'm curious about your use case. What requires you to use it in the first place?
I'm using a model to update specific fields just like you said, what the `ColumnValueMapper` intended. 

I was using a custom model first. Then I notice most of the fields were no need to update at all. So, I create a 'Mapper' to fix this, and not distinguish  insert and update properly.



Thanks for the quick review. I'm looking forward the DAO rework.


> @sospartan Yep, that part of the code will definitely will need to change (I've just noticed that we are using the `id` in the `update()`). I'll have it in mind when reworking the base Dao methods related to the custom insertion id.

I've noticed that too. :)

ganigeorgiev:
On second thought, your approach to use directly the id in the `Dao.create` maybe is OK for this case because the base model interface operates only on a single primary key `id` (aka. has methods like GetId, RefreshId, etc), so we can make an exception for the `id` field to avoid manually parsing the model tags (the underlying db lib ozzo-dbx also uses the ID/id as a default primary key). **But I'll think a little more on this once I start working on the user provided id feature.**

Anyway, unless your db model stores large blobs or has a lot of fields with indexes, I wouldn't bother too much with trying to optimize the update query and will update all model fields at once (it is just less error-prone this way, since you don't have to worry that you'll forget to map some field).
If you really want to update only a specific set of model fields, then you could use directly the db abstraction's `Update()` method, something like:
```
dao.DB().Model(myModel).Update("field1", "field2")
```

The `ColumnValueMapper` was really created only because of the `Record` model as a workaround when your model fields are dynamic and/or when your model data doesn't directly map 1:1 to its related table schema.

sospartan:
I'll try the `Update()` method.  Never use dbx before, seems interesting.

PR #282, opened 2022-08-09
(feature): ip checking fallback for logs

theMackabu:
allow ip to be logged when using reverse proxies

### DIFF
--- a/apis/middlewares.go
+++ b/apis/middlewares.go
@@ -210,7 +210,7 @@ func LoadCollectionContext(app core.App) echo.MiddlewareFunc {
 func ReadUserIP(r *http.Request) string {
 	 IPAddress := r.Header.Get("CF-Connecting-IP")
 	 if IPAddress == "" {
-		  IPAddress := r.Header.Get("X-Real-Ip")
+		  IPAddress = r.Header.Get("X-Real-Ip")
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
This is a nice addition, but I need to think more on it.

In any case:
- Could be simplified by listing all ips in a slice (eg. `[]string{r.Header.Get("CF-Connecting-IP"), ...}`) and return the first non empty value from it
- Need a simple unit test
- Need a comment indicating that this is not safe to use if you are not behind a trusted reverse proxy

For the logs it may be a good idea to store both IPs, but I'll think more on it during the weekend.

ganigeorgiev:
@theMackabu To avoid adding a new configuration option where developers will have to check whether they want to use a trusted proxy IP or the remote address, I think it will be better to just add both ips in the `Request` model.
I'll close the PR but I'll consider implementing the necessary changes to the logs in the next v0.5.0 or v0.6.0 release.

ganigeorgiev:
I've decided to store both IPs as `userIp` and `remoteIp` (previously just `ip`).

This is implemented in `master`. Will be available in the next v0.5.0 release (the JavaScript and Dart DTOs will be also updated to reflect the change).

PR #288, opened 2022-08-10
Communicate default thumb size

### LINKED ISSUE #287: Communicate default thumbnail size
The default thumbnail size (100x100) is not displayed in the UI when you edit a file field:
Screenshot from 2022-08-10 18-46-46
Maybe this could be nice to communicate to the user? For me, it was a bit confusing to see that I had thumbnails despite this field being empty and also it required some digging in the source code to find where this was actually [set](https://github.com/pocketbase/pocketbase/blob/3d07f0211dc74710affd9154f61728d77cfb6f4c/apis/file.go#L13) and [used](https://github.com/pocketbase/pocketbase/blob/3d07f0211dc74710affd9154f61728d77cfb6f4c/apis/file.go#L63):

Let me know what you think :slightly_smiling_face:

ollema:
Fixes https://github.com/pocketbase/pocketbase/issues/287 (if you think it needs a fix to begin with!)

### DIFF
--- a/ui/src/components/collections/schema/FileOptions.svelte
+++ b/ui/src/components/collections/schema/FileOptions.svelte
@@ -126,7 +126,7 @@
                 <i
                     class="ri-information-line link-hint"
                     use:tooltip={{
-                        text: "List of thumb sizes for image files. The thumbs will be generated lazily on first access.",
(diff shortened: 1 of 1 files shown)


ollema:
**Before:**
Screenshot from 2022-08-10 18-46-46

---

**After:**
Screenshot from 2022-08-10 19-06-51

ganigeorgiev:
Thanks! Will be included in the next release (v0.4.0).

PR #333, opened 2022-08-20
Orden las colecciones alfabeticamente

zcastle:
Orden las colecciones alfabeticamente

### DIFF
--- a/apis/collection.go
+++ b/apis/collection.go
@@ -37,6 +37,7 @@ func (api *collectionApi) list(c echo.Context) error {
 
 	result, err := search.NewProvider(fieldResolver).
 		Query(api.app.Dao().CollectionQuery()).
+		Sort(search.ParseSortFromString("name")).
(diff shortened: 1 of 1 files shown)


irmhonde:
Hi @zcastle. Is it strictly necessary to hardcode an alphabetical sort since it is already possible through client sdks and REST as such?

dart-sdk: `requires Admin Token`
```dart
final collections = await client.collections.getFullList(batch: 200, sort: '-name');
```

js-sdk: `requires Admin Token`
```js
const collections = await client.collections.getFullList(200 /* batch size */, { sort: '-name' });
```

REST: `requires Admin Token`
```
http://<ip>:8090/api/collections?sort=-name
```

ganigeorgiev:
As @irmhonde mentioned, if you need to sort the collections the recommended way to do it is to use the `sort` query parameter.

The Admin UI sorts the collection by the `created` date. I find this to be more easy to work it but I'll consider in future version to add an option for users to change it.

PR #372, opened 2022-09-02
fix install commond

dxsdyhm:
fix install commond

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ your own custom app specific business logic and still have a single portable exe
 
 ```sh
 # go 1.18+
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
@dxsdyhm Hi, thanks for the PR but I don't think it's necessary.

`go get` is used because the command is expected to add PocketBase as any other Go library to your go.mod.

You don't need to have or use Go to install the prebuilt executable. If you want to use the prebuilt binary, you can download it directly from the [releases page](https://github.com/pocketbase/pocketbase/releases).

PR #373, opened 2022-09-02
chore: add star history chart to README

boojack:


ganigeorgiev:
@boojack Thanks for the PR, but for now I don't think it is needed and will just "clutter" the README without bringing too much value (_in addition, I'm also planning to add more info and examples in the future_).

Issue #376, opened 2022-09-02
Merge users and profiles (request for feedback)


ganigeorgiev:
_This was partially discussed in https://github.com/pocketbase/pocketbase/discussions/225#discussioncomment-3511148 and other similar older discussions._

The current approach of having 2 separate entities (users and profiles) to manage the application users is not very convenient and has several drawbacks:

- It is confusing (and unnecessary) to have both `user` and `relation` type fields.

- Setting a profile field value during user create is too verbose and nonintuitive. Currently the process requires sending multiple non-transactional requests:
  ```js
  // create the user model
  await client.users.create({
    email: "test@example.com",
    password: "123456",
    passwordConfirm: "123456",
  });

  // authorize as the previously created user
  await client.users.authViaEmail('test@example.com', '123456');

  // update the user profile associated to the previously created user
  await client.records.update('profiles', client.authStore.model.id, {
     name: "test",
     anotherCustomField: "Lorem ipsum...",
  });
  ```

- The `profiles` collection has `create` and `delete` API rules, but they are not really useful. Furthermore, because an empty profile is created automatically on user create, the `profiles` create API rule is not actually triggered (_unless the profile record was deleted and then created manually_).

- Because the users api doesn't have API rules, currently everyone is allowed to create a new user.

- In addition to the above, there also no easy support for full "manager-subordinate" users management.
  While one user could be allowed to edit the profile data of another user, currently it is not possible, without using PocketBase as framework, to allow for example one user (eg. _manager_) to completely manage or delete another user (eg. _subordinate_).

**Considering the above shortcomings, I propose to merge the `user` models and the `profiles` collection into a new single system collection called `users`.**

This basically means that:

- We will remove the `user` type field and migrate it existing usage to a `relation` type field (pointing to the new `users` collection).

- We will be able to use the new collection's API rules to better control who can access and perform each crud user operation.

- We will be able to create/update a user with just one request, something like:
  ```js
  await client.users.create({
    email: "test@example.com",
    password: "123456",
    passwordConfirm: "123456",
    name: "test",
    anotherCustomField: "Lorem ipsum...",
  });
  ```

- Because of the specifics and the edge cases related to the system fields (email, verified, password) we will have to modify the default behavior of the `record_upsert` form.

While I'll try to automate as much as possible the required DB changes, **this unfortunately will not be fully backward compatible**.
Of course, there will be a detailed migration guide, but still users will have to update manually their existing custom app code.

[...]

PR #392, opened 2022-09-05
Fix auth issue when creating new records

evanlurvey:
When I was trying to design a role members table for a multi tenant thing I am working on I noticed an issue where if the incoming create request had the permissions I was looking for it would allow it. 
This PR adjusts the join to use a subquery which excludes the id of the test record. 

Not sure if this is structurally where something like this should exist. I am happy to make any changes or refactors.

### DIFF
--- a/apis/base.go
+++ b/apis/base.go
@@ -134,7 +134,7 @@ func StaticDirectoryHandler(fileSystem fs.FS, disablePathUnescaping bool) echo.H
 }
 
 // bindStaticAdminUI registers the endpoints that serves the static admin UI.
-func bindStaticAdminUI(app core.App, e *echo.Echo) error {
+func bindStaticAdminUI(app core.App, e *echo.Echo) {
 	// redirect to trailing slash to ensure that relative urls will still work properly
 	e.GET(
 		strings.TrimRight(trailedAdminPath, "/"),
@@ -152,7 +152,7 @@ func bindStaticAdminUI(app core.App, e *echo.Echo) error {
 		middleware.Gzip(),
 	)
 
-	return nil
+	return
 }
 
 const totalAdminsCacheKey = "totalAdmins"
--- a/apis/record.go
+++ b/apis/record.go
@@ -177,19 +177,20 @@ func (api *recordApi) create(c echo.Context) error {
 
 	// temporary save the record and check it against the create rule
 	if admin == nil && collection.CreateRule != nil && *collection.CreateRule != "" {
+		testRecord := models.NewRecord(collection)
+		testForm := forms.NewRecordUpsert(api.app, testRecord)
+
 		ruleFunc := func(q *dbx.SelectQuery) error {
-			resolver := resolvers.NewRecordFieldResolver(api.app.Dao(), collection, requestData)
+			resolver := resolvers.NewRecordFieldResolver(api.app.Dao(), collection, requestData, testRecord)
 			expr, err := search.FilterData(*collection.CreateRule).BuildExpr(resolver)
 			if err != nil {
 				return err
 			}
-			resolver.UpdateQuery(q)
+			_ = resolver.UpdateQuery(q)
 			q.AndWhere(expr)
 			return nil
 		}
 
-		testRecord := models.NewRecord(collection)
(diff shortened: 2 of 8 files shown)


ganigeorgiev:
@evanlurvey Could you elaborate what exactly is the issue?

The idea of the existing check inside `DrySubmit` is to verify that the transaction created record pass the Collection Create API rule.

evanlurvey:
So say we have a query like this one 
```
@collection.tenant_member.role = "admin" && @collection.tenant_member.member = @request.user.id &&
@collection.tenant_member.tenant.id = tenant.id
```
The schema is something like this:
tenant:
  - name
tenant_member:
  - user
  - tenant
  - role
```

To me that query reads if I find an existing tenant member with the role admin, using the current request user, as long as the tenant member record I want to create matches a tenant I am an admin in then create the record. 

So with an empty database if I send a request like POST `/api/collections/tenant_member/records` with a payload of 
```
{
    "tenant": "<any tenant id>",
    "member": "<my user id>",
    "role": "admin"
} 
```
I would expect to get a permission denied if there wasn't already an existing admin role associated with my account and that tenant. But currently it will allow it. This is because when it does the dry run it actually inserts a record then does all the queries to compare. Since it was not excluding the current record we could add any qualifying data needed to make the insert happen. 

This fix simply excludes the test record that gets inserted to check permissions from the @collection query.

ganigeorgiev:
I think there is some misunderstanding how the API rules work.

To me it doesn't make sense to add a subquery to exclude the created record - if we wanted this behavior there wouldn't be a need for a `DrySubmit` in the first place.

To be honest, I'm not sure that I understand what actually you are trying to do. The following 2 statements:
"_as long as the tenant member record I want to create matches a tenant I am an admin in then create the record_" and "_I would expect to get a permission denied if there wasn't already an existing admin role associated with my account and that tenant._" seems to contradict with each other.

`@collection.*` filter is usually intented for querying other indirect collections, aka. collections to which you don't have a direct relation field. If you apply the example `@collection.*` filter rule on the `tenant_member` it will do practically a left join with itself.

Since I'm already doing a refactoring to the users management in https://github.com/pocketbase/pocketbase/issues/376, I'll think how to "normalize" the create rule behavior because it indeed differs from the `list`, `view`, `update`, `delete` rules as mentioned in https://github.com/pocketbase/pocketbase/discussions/225#discussioncomment-3512136 (or at least to better document the nuances around it).

evanlurvey:
I agree it could be done without the dry submit. This was a way to make it work with that mechanism. 

Without something like this in place a privilege escalation can happen with the situation given. 

I guess the same could happen by adding `@collection.tenant_member.id != id` Which would get the collection join to exclude the dry run record. 

I haven’t tested it yet. But that should do the same and could then be a documentation update instead of a code change.

evanlurvey:
For further clarification on my scenario:

When a user wants to add someone to their tenant (assume some B2B application) I would only expect that add to go through if they were already an admin. However with the test record already “in” the collection at time of query, if the id of the test record isn’t excluded the user could create a record that makes them the admin of any tenant.

ganigeorgiev:
> When a user wants to add someone to their tenant (assume some B2B application) I would only expect that add to go through if they were already an admin. However with the test record already “in” the collection at time of query, if the id of the test record isn’t excluded the user could create a record that makes them the admin of any tenant.

The create API rule runs after the record is "inserted" in the database, otherwise you wouldn't have access to the record fields at all. If you want to restrict the submitted user data, you could:
- use [`OnRecordBeforeCreateRequest` event hook](https://pocketbase.io/docs/event-hooks/#record-hooks) (if you are using PocketBase as framework)
- OR check the submitted request data in the rule by don't allowing role "admin" assigment, eg. `@request.data.role != 'admin'`

ganigeorgiev:
@evanlurvey In any case, the proposed changes in https://github.com/pocketbase/pocketbase/issues/376 should simplify the users management (you will even have an options for different auth models on app level).

If you have any other ideas or suggestions how to improve the current behavior, please share your feedback in https://github.com/pocketbase/pocketbase/issues/376.

PR #411, opened 2022-09-11
DRAFT: Instagram auth provider

Nymuxyzo:
I added Instagram as a new auth provider to pocketbase. Unfortunately, I don't have a Facebook or Instagram account, so I can't fully test it. Maybe someone could thankfully help me.

Source:
https://developers.facebook.com/docs/instagram-basic-display-api/reference/oauth-authorize

I don't know if I have to do a `npm run build` because of ui/src/providers.js


ganigeorgiev:
> I don't know if I have to do a npm run build because of ui/src/providers.js

Yes, if you make changes to the ui, it is necessary also to run `npm run build` to make sure that the ui/dist contains your changes. Threre is also a small contribution guide here that you could check - https://github.com/pocketbase/pocketbase/blob/master/CONTRIBUTING.md

_Maybe it will be better to put this on aside for a while because there are expected to be major changes to the tests because of the [users management refactoring](https://github.com/pocketbase/pocketbase/issues/376)._

ganigeorgiev:
@andre-gitdev I'll close the PR for now, since I'm planning some changes to the tests and other internals and it will be hard to keep your branch in sync. Similar to  PR for the Twitter provider, we can refer to this implementation after I manage to merge the changes in #376 (after all, the most important part is the provider struct itself and there will be no changes to that).

Issue #421, opened 2022-09-13
Serve fallback file when requested file does not exist


RoenLie:
Hi!

Is there a standardized way to redirect back to /index.html whenever a requested file does not exist?

Reason why I am asking is because I am serving a SPA application, which handles its own routing.

I would like pocketbase to allow the SPA to handle routing requests that don't match existing files in the public directory.

At the moment I am using the example for serving public files, found in this repo.

app.OnBeforeServe().Add(func(e` *core.ServeEvent) error {
  // serves static files from the provided public dir (if exists)
  subFs := echo.MustSubFS(e.Router.Filesystem, "pb_public")
  e.Router.GET("/*",  apis.StaticDirectoryHandler(subFs, false))

return nil
})

Thank you!

RoenLie:
I got it working by copying the apis.StaticDirectoryHandeler method and adding a check if the file exists.
But if there is an intended way to do this, it would be great.
.net exposes a method for this spesific purpose.
```endpoints.MapFallbackToFile("/index.html");```

```
func StaticDirectoryHandlerCustom(fileSystem fs.FS, disablePathUnescaping bool) echo.HandlerFunc {
	return func(c echo.Context) error {
		p := c.PathParam("*")
		if !disablePathUnescaping { // when router is already unescaping we do not want to do is twice
			tmpPath, err := url.PathUnescape(p)
			if err != nil {
				return fmt.Errorf("failed to unescape path variable: %w", err)
			}
			p = tmpPath
		}

		// fs.FS.Open() already assumes that file names are relative to FS root path and considers name with prefix `/` as invalid
		name := filepath.ToSlash(filepath.Clean(strings.TrimPrefix(p, "/")))

                // Checks if file exists, redirects to index if it does not.
		_, err := fs.Stat(fileSystem, name)
		if err != nil {

			return c.FileFS("index.html", fileSystem)
		}

		return c.FileFS(name, fileSystem)
	}
}
```

ganigeorgiev:
@RoenLie Yes, overwritting the StaticDirectoryHandler is one way to do it. Another possible workaround could be to "extend" the global error handler:

```go
package main

import (
	"log"
	"net/http"

	"github.com/labstack/echo/v5"
	"github.com/pocketbase/pocketbase"
	"github.com/pocketbase/pocketbase/apis"
	"github.com/pocketbase/pocketbase/core"
)

func main() {
	app := pocketbase.New()

[...]

PR #554, opened 2022-09-19
add spa flag to serve static files in spa mode

mstip:
Added --spa as a flag to serve static files in spa mode.
This is useful if you use pocketbase pb_public dir to serve react, angular, vue or whatever spa framework you like. 
If in spa mode, the StaticDirectoryHandler falls back to index.html when the file is not found and contains no dot. 
For example, /deep/deep/link would fall back to index.html but /unknown.png would still return 404.

### DIFF
--- a/apis/base.go
+++ b/apis/base.go
@@ -115,7 +115,7 @@ func InitApi(app core.App) (*echo.Echo, error) {
 // but without the directory redirect which conflicts with RemoveTrailingSlash middleware.
 //
 // @see https://github.com/labstack/echo/issues/2211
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
This was recently discussed in https://github.com/pocketbase/pocketbase/issues/421#issuecomment-1246018247 and it is planned after the refactoring in #376.

The idea to ignore files is a good one, but may not be reliable (you can have files without extension).

I'll close the issue for now, but I have it in my todo and we can always refer to it after the refactoring (please note that there will be a lot of internal changes and it's better new PRs to wait).

Issue #340, opened 2022-08-22
Feature Request: Twitch Auth Provider


KopiNinja:
Your project is fantastic! 
I'm not sure how difficult it is to implement Twitch auth, but it would be awesome to use with Twitch chat bot.

ganigeorgiev:
Twitch has [OAuth2 api](https://dev.twitch.tv/docs/authentication) with [scopes to return the user email](https://dev.twitch.tv/docs/authentication/scopes) but unfortunately they don't seem to support the PKCE flow.

There is an existing feature request in the Twitch uservoice forum to add support for that - https://twitch.uservoice.com/forums/310213-developers?query=pkce

For now I'll close the issue because as it is now, this could work securely only in the browser, but depending on the users demand for this provider, I may reconsider it.

PR #555, opened 2022-09-19
Added Twitch OAuth2 provider

AdrianGonz97:
Hi! This is a great project and I would love to replace my reliance on Supabase for it. In order to do so, I needed to add Twitch as an OAuth provider.

Most of the new code is reused from the other existing providers. A new method for fetching the user data was created as Twitch requires a "Client-Id" header for that request.

I'm aware that issue #340 was closed due to the lack of PKCE support so if that's still a major issue for implementing this provider, I'd understand the closing of this PR.


ganigeorgiev:
Hi! Thanks for the contribution but please note that PR for new features will be reviewed and merged only after the refactoring in #376. There will be a lot of changes to the internals, especially to the tests, so please be patient.

I'll close the PR for now to avoid spamming with notifications for conflicting changes, but we can always refer to it later. I'll add it to my todo to review the proposed changes after #376.

PR #585, opened 2022-09-20
Respect the EXIF auto orientation tag.

a10d:
When uploading an image from a mobile device, the generated thumbnails are not rotated according to the EXIF tag. This results in thumbnails being potentially generated in the wrong orientation.

The [imaging package supports reading the EXIF tag](
https://github.com/disintegration/imaging#incorrect-image-orientation-after-processing-eg-an-image-appears-rotated-after-resizing), so adding support for it is a simple fix.


ganigeorgiev:
Thanks! It will be included in the next bugfix release sometime later this week.

PR #643, opened 2022-09-26
Add a Show/Hide password option

TylerSustare:
# Adds the ability to show and hide the password for login and "installer" pages in PocketBase.

I wanted this feature when I created my first PocketBase app, so I thought I could help out the PB project 😄 

## What should reviewers focus on?
The main content to review is in commit #2 where the logic was added to use a `password` or `text` input type. I decided to add the eye that toggles seeing the password to the label, rather than the input field itself. This was for two reasons. The main reason is, I don't have a whole lot of time for this feature, and this was the quickest way. The other reason I went with this approach is: it is very minimalistic, like PocketBase, and didn't clash with the existing design.

I had to get rid of `bind:value={password}` when using `type={show ? "text" : "password"}` on the input. The  'type' attribute cannot be dynamic if the input uses two-way binding. It resulted in a `svelte(invalid-type)` error. To get around this I made small setter functions and used `on:change`. This feels much more the React way, than the Svelte way. However, [according to this comment](https://github.com/sveltejs/svelte/issues/3921#issuecomment-553651625), Rich Harris thinks it's OK 😅 

## Preview of the changes

Screen Shot 2022-09-25 at 9 30 45 PM


ganigeorgiev:
This is a nice suggestion, but it'll need to be implemented slightly differently.

There is no need for external svgs. The icon font that we are currently using ([remixicon](https://remixicon.com/)) already has suitable icons.

Additionally, to avoid duplication we can also move the logic into a separate `PasswordInput.svelte` component.

To be honest, I'm not sure that I like the position of the toggle-icon because it's not very clear that it is "actionable".
This icon position currently is mainly used as meta to display the type of the different collection fields. Actionable input controls in PocketBase are usually placed on the right side of the input (see for example the password field in the SMTP mail settings), but now I see that it may collide with some browser extensions as shown in your screenshot (I guess this is 1pass?) so I may need to reconsider this.

_I'll close the PR for now, but I'll add it to my todo and will consider implementing it either as part of the refactoring in #376 or the release after that._

TylerSustare:
@ganigeorgiev Those are perfectly valid points 😃 
The right-side action definitely makes sense. It's a standard at this point.

Thanks for looking into this, I hope this PR was helpful as a starting point at least 🎉

PR #701, opened 2022-10-02
PBScript integration discussion

benallfree:
Hi @ganigeorgiev,

I have started a project https://github.com/benallfree/pbscript/ to add Javascript server-side support to PocketBase via goja and as discussed here https://github.com/pocketbase/pocketbase/discussions/498#discussioncomment-3673686

I'm not sure if this is a frequently requested feature or if you have any interest in someday integrating JS function support, but I wanted to start this PR as a place to discuss.

In this PR, I demonstrate how to instantiate a PB instance with JS function support. This is very alpha stuff, but I can get Go and JS talking to each other and even respond to API routes and basic PB event hooks.

### DIFF
--- a/examples/base/main.go
+++ b/examples/base/main.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/benallfree/pbscript/modules/pbscript/engine"
 	"github.com/pocketbase/pocketbase"
 	"github.com/pocketbase/pocketbase/apis"
 	"github.com/pocketbase/pocketbase/core"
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
@benallfree Thank you for working on this, but as mentioned in previous issues, for now I don't plan adding serverless functions support as part of the core.

In the future we may add a "Workflow builder" where something similar could be added dynamically as a plugin/block but for now this is left out of the scope for v1 because I don't think I'll have the time to work properly on this feature.

In any case, you've done a great job making it pluggable so I'm sure that it would be helpful for those not familiar with Go.

PR #759, opened 2022-10-12
Views

iskaa02:


iskaa02:
# SQL views
Add way to run raw SQL queries through views. this is useful for complex queries and without editing the go source, the api work like collections
### Creating view
POST request to `pocketbase.url/api/views`
with payload `name:string, sql:string, listRule:string|null`
example
```json
{
  "name":"get_posts",
  "sql":"select * from post",
  "listRule":null
}
```
response example 
```json
{
  "id": "q8cxrwlsfrsvmya",
  "created": "2022-10-11 23:10:57.456",
  "updated": "2022-10-11 23:22:10.849",
  "sql": "select posts.* from posts",
  "name": "view_demo",
  "listRule": null,
  "schema":[{
    "system": false,
    "id": "vvdrd1qz",
    "name": "created",
    "type": "text",
    "required": false,
    "unique": false,
    "options": {
	"min": null,
	"max": null,
	"pattern": ""
         }
      }
  ]
}
```
delete, update, records work the same as collections

SQL field is validated before inserting, rules for a valid SQL:
- working select query
- not affecting any rows
- no semi colons, everything after the first semi colon is removed
## Api Rules
there's only one rule which is the list rule and it work the same as collection records.
 when having a duplicated column name SQLite by default prefixing the duplicated column name with ':1', a colon is an invalid api rule character, and it's confusing to use have two columns the same name, so when the admin want to use the duplicated column it should be renamed in the query.

## Schema
the schema is generated after the view is created.
json, number and boolean fields are parsed but other types like file,date,..etc use the same underlying sql type 'TEXT' we can't infer those types i thought about doing a parsing when records are requested and parsing each record by a regex or something but that might not be the best solution so i left them as normal strings.

## On collection change
when a collection change, field type or name change, more or less fields the view do not update it will show the old view schema meaning that any new field will not be included in the query, the admin should update the view manually by sending a patch request without changing any thing this will trigger the view to re-generate the schema.


 I try to follow the package way of doing things as much as possible so there's hook for views [create update delete]  and also for the request list.

## JS sdk
I modified the JS SDK to work with views features, note there's one test that field i could not figure out why, but everything is working just fine.

## UI
I build the UI for interacting with views it includes SQL editor with completion collections, rule field with completion (since the schema is generated after the view is created the completion in the rule field is will not work properly unless the you create the view first and then edit the rule)
there's a bug in the UI i could not figure out, I'm not that familiar with svelte sorting fields in the views page don't work for some reason, i tried curl with sorting it work just fine.
### Running the UI
The ui uses a modified JS sdk so if you want to run it clone the [sdk repo](https://github.com/iskaa02/js-sdk) first, cd into the ui directory then run `$ npm install path/to/js-sdk`

## Security
Exposing SQL can be dangerous, i tried to do as many check i can to make sure the provided SQL is safe read only and not affecting any data, I used the `total_changes()` SQLite function to check if the query is affecting any rows, inside a transaction i run the `total_changes()` function before and after executing the query, then i rollback, after checking if the query is safe the view is created, i tried opening a new database connection with parameter `mode=ro` (read-only mode) but for some reason that didn't work as queries affect rows without any error being returned.

###

My Use case for pocketbase require executing complex SQL queries, i built a basic implementation for myself first then i thought that someone might need this feature too.

ganigeorgiev:
Thank you for working on this, but currently PR for new features are temporary frozen due to the refactoring in #376.

Additionally, I think something like this would be better implemented as a Collection type (collection types will be available soon with v0.8), but for now I don't plan merging it in PocketBase. We ca always refer to this implementation and take bits of it when we decide to implement it.

In any case, please refrain investing too much time in big feature PRs without first outlining some implementation details in a discussion/issue. I want to keep PocketBase as simple as possible. Every new feature carries an increase in the project complexity with it and need to be carefully considered (adding new features is "easy", but removing them is harder and even impossible after v1.0.0).

Ideally, features like this one would be better implemented as "plugins", aka. a separate package that accept the `core.App` interface and uses its hooks to register the wanted custom behavior. @benallfree has done something similar in https://github.com/benallfree/pbscript. And depending on the users demand and how used this "plugin" is, we can later directly import it as part of the core package. _This approach has one main limitation though and that is that currently there is no easy way to dynamically load new frontend views into the existing Admin UI but I'll think a little more on this when researching the "Collection triggers/Workflow builder" in the future_.

benallfree:
@iskaa02 This is cool and also I agree with @ganigeorgiev about keeping the core small. I say this even as I have done a lot of work extending PocketBase like you have.

If you can reorganize this into a PocketBase plugin like I did with https://github.com/benallfree/pbscript, I would be happy to add it to [awesome-pocketbase](https://github.com/benallfree/awesome-pocketbase) to increase visibility of your project. I will also consider supporting in at pockethost.io which uses a custom build running PBScript already.

As you already found, the admin and client SDK don't really have a plugin architecture like the PocketBase core does. I've settled on wrapping the `pocketbase` JS client in a custom function that returns higher-level application-specific helpers. So instead of modifying the JS SDK, you could publish a project like `pocketbase-client-views` or something that would work like this:

```ts
import makeViewHelpers from 'pocketbase-client-views'
const client = new PocketBase(...)
const viewHelpers = makeViewHelpers(client)
```

benallfree:
@iskaa02 In my opinion it would definitely be a worthwhile effort if you wanted to create some kind of hook/extension API for the admin. Here's my first thought about it:

1. Admin UI calls a PocketBase REST API endpoint, maybe `/admin-plugins`
2. The REST endpoint fires a PocketBase hook to gather admin plugin JS bundles. Any listeners can respond to the event by adding a bundle, maybe there is an `e.AddAdminJsBundle(js string)` function passed in the event context.
3. All admin bundles must export a default function `module.exports = (adminContext: AdminContext)=>{...}` that is passed some kind of `AdminContext` from the main admin.

The `AdminContext` would have various methods like `getSections()` and `addSection()`

What do you think of that idea? I would be interested in collaborating on that with you and @ganigeorgiev because it would be a useful thing for PBScript as well.

benallfree:
just to expand on this a bit more, here's what the plugin might look like if it were written in React:

```ts
const MyIcon = ()=><div>icon</div>
const MyPage = ()=><div>page</div>

const myPlugin : AdminUiPlugin = (ctx:AdminUiContext)=>{
   ctx.addSection( {
      name: 'My Section'
      iconRender: ( iconContainerEl: HTMLDivElement )=>{
         const root = ReactDOM.createRoot(iconContainerEl);
         root.render(e(MyIcon));      
      }
      render: ( pageContainerEl: HTMLDivElement )=>{
         const root = ReactDOM.createRoot(pageContainerEl);
         root.render(e(MyPage));      
      }
}
export = myPlugin
```
image

ganigeorgiev:
Yep, something like the above could be implemented, but there are also other things that will have to do before that, like how to expose the inner SPA router, how to provide access to some of the global app stores, export the current ui components in a dedicated library so that the styles can be reused and have a more unified style, how external js/css will be loaded, etc.
Let's not rush it. I have it in my todo and it will research it in more details when I start working on the "Workflow builder" feature.

ganigeorgiev:
_(Just for info in case anyone stumbled on this PR)_

Experimental "View" collection type was added in the `develop` branch and will be available with the next v0.13.0 release (most likely next week or the week following it).

Internally the implementation slightly differs from this PR, but the general idea is the same - to allow users to perform a custom `SELECT` query while preserving the same collection APIs (including list and view access rules, relation expansion, filters, sort, etc.).

The only limitation is that the query must have a unique `id` column (this is to preserve the `getOne()` api). The base models `created` and `updated` fields are optional for the `view` collections and will not be returned in the response if such columns don't exist or are empty.

The fields detection ended up a little more complicated than I initially anticipated, because `table_info` doesn't return enough information about the columns to properly map them to their related field definitions, so a very rudimentary query tokenizer and parser was implemented. It supports:

- column and table aliases (including the sqlite supported quotes)

- auto `relation` fields associations, eg:
  ```sql
  -- bid will be a "relation" field
  select a.id, b.id as bid
  from a
  left join b
  ```

- file fields proxying (up to 5 nested relations, eg. view -> view -> base), allowing access to view file colums as seamless as possible without copying the originals

- some special normalizations for numeric and cast expressions (eg. `count()`, `total()`, `cast(foo as real)`, etc.)


There are still some minor polishing things that need to be done (and strangely one of the tests fails when executed with the CGO driver for some reason that I'll have to investigate), but if anyone want to test it, feel free to checkout the [`develop` branch](https://github.com/pocketbase/pocketbase/tree/develop).

Issue #746, opened 2022-10-08
Feature Request: Adding Microsoft AD Oauth2


WouterVandeWiele:
Hi all,
Would it be possible to add Microsoft AD as an authentication option? Not sure if it is on the roadmap or a wish list.

I see there is a similar issue open for Apple Signin (#202). Past few days I had some time and was looking through the instructions provided in that thread and came up with this: https://github.com/WouterVandeWiele/pocketbase/commit/99d55f3e25a55673c4469d05e596319a934f74a2

Unfortunately I could not test it yet on a "Microsoft 365 Developer Program - E5 subscription" due to a [technical issue.](https://learn.microsoft.com/en-us/answers/questions/1035772/unable-to-setup-microsoft-365-developer-program-e5-1.html) :disappointed:

ganigeorgiev:
This was requested previously on the support mail too.

I'll consider it for a future a release, but for now it is a low priority.

astrojarred:
This would also be super useful for my use case. @WouterVandeWiele if you don't resolve the issue with the developer program in the next week or so let me know and I could try to help test.

WouterVandeWiele:
@astrojarred thanks for the offer, fortunately the problem with the test tenants was solved.

So a small update: I was able to create the a test instance, register an app, create a client secret and located the auth and token endpoints to use.

Next I took the example from the documentation "[Auth via OAuth2](https://pocketbase.io/docs/manage-users/#auth-via-oauth2)" and served them locally.

- The index.html page works fine, it finds the Microsoft auth provider and when clicked redirects to the auth url
- Going through the login flow, and at the end I get redirected to the redirect.html page
- Initially on the redirect.html page, it says `Authenticating...` next the promise returns:
```
Failed to exchange code.
ClientResponseError 400: Failed to authenticate.
```
In the logs it displays: [`Missing required ExternalAuth field`](https://github.com/pocketbase/pocketbase/blob/master/daos/external_auth.go#L78), and after some investigation I saw that the ProviderId is empty.

This ProviderId seems to originating from the SQLite DB `_externalAuths` table, but when I try to inspect this table it is empty.

To try and debug this issue, I've also added google oauth. When I first login via google I can see this log message:
```
INSERT INTO `_externalAuths` (`created`, `id`, `provider`, `providerId`, `updated`, `userId`) VALUES ...
```
But nothing similar shows up when authenticating with the Microsoft user ☹️ 

[...]

PR #873, opened 2022-10-30
Rc

WouterVandeWiele:
Adding Microsoft Azure AD authentication.  #746

` authentmake test` failures:
```shell
...
=== RUN   TestAuthProviderConfigSetupProvider
--- PASS: TestAuthProviderConfigSetupProvider (0.00s)
FAIL
coverage: 76.4% of statements
FAIL    github.com/pocketbase/pocketbase/core   0.043s
...
=== RUN   TestPointer
--- PASS: TestPointer (0.00s)
PASS
coverage: 97.5% of statements
ok      github.com/pocketbase/pocketbase/tools/types    0.002s  coverage: 97.5% of statements
?       github.com/pocketbase/pocketbase/ui     [no test files]
FAIL
make: *** [Makefile:5: test] Error 1
```


core/settings.go
```diff
@@ -33,12 +33,13 @@ type Settings struct {
...
-	GitlabAuth   AuthProviderConfig `form:"gitlabAuth" json:"gitlabAuth"`
-	DiscordAuth  AuthProviderConfig `form:"discordAuth" json:"discordAuth"`
-	TwitterAuth  AuthProviderConfig `form:"twitterAuth" json:"twitterAuth"`
+	GoogleAuth      AuthProviderConfig `form:"googleAuth" json:"googleAuth"`
+	FacebookAuth    AuthProviderConfig `form:"facebookAuth" json:"facebookAuth"`
+	GithubAuth      AuthProviderConfig `form:"githubAuth" json:"githubAuth"`
+	GitlabAuth      AuthProviderConfig `form:"gitlabAuth" json:"gitlabAuth"`
+	DiscordAuth     AuthProviderConfig `form:"discordAuth" json:"discordAuth"`
+	TwitterAuth     AuthProviderConfig `form:"twitterAuth" json:"twitterAuth"`
+	MicrosoftAdAuth AuthProviderConfig `from:"microsoftAdAuth" json:"microsoftAdAuth"`
```
ganigeorgiev, line 42: I think we can remove the "Ad" suffix from the provider name and all other places, aka. let's keep it just "MicrosoftAuth".

ganigeorgiev (changes requested):

core/settings.go
ollema, line 42: is `from:` a typo here? all other providers user `form:`

WouterVandeWiele: oops. Should be fixed now.

WouterVandeWiele:
Changed the microsoftAuth names to azureAuth.

Also reordered the expected result in `core/settings_test.go` - `TestNamedAuthProviderConfigs`. Now the test failures from `make test`, mentioned in earlier, are gone.

ganigeorgiev:
@WouterVandeWiele Thank you. I've merged and squashed the PR locally in https://github.com/pocketbase/pocketbase/commit/5298543ce486b0e4fe53dc1fc9a38b02313e22c5.

At the end, I've decided to keep the provider name "Microsoft" for simplicity. Other than "Windows's Live ID" I couldn't find any other OAuth2 services (and it seems that it is a legacy as far as I understand reading https://techcommunity.microsoft.com/t5/microsoft-entra-azure-ad-blog/cleaning-up-the-azuread-and-microsoft-account-overlap/ba-p/245105 and other related resources). Additionally Microsoft Azure AD seems to support both cases - Microsoft personal accounts as well as Azure Active Directory accounts (Firebase also seems to use only "Microsoft" as label in their UI and documentations https://firebase.google.com/docs/auth/android/microsoft-oauth).

I've also decided to add support for custom provider components loaded through the `provider.js` config file, which should result in better UX since now we can show specialized help-blocks or hide unnecessary fields (eg. the user api endpoint for the microsoft provider).

I've tested the integration with a developer account and it seems to work fine. If you stumble on something or don't agree with the changes, feel free to let me know I'll revisit them.

WouterVandeWiele:
I concur on the naming.

The `optionsComponent` is also a great addition. 👍 
It's more customizable then the old way while keeping a clean interface IMHO.

Great work, thanks for adding the feature.

PR #899, opened 2022-11-02
Implement Apple Oauth2 Flow

### LINKED ISSUE #202: [Need help with testing] Feature request: Apple Signin
Apple Signin is also a great option. Do we have plan for it? Or could I start a draft PR for it?

n1xx1:
Stuff missing:
- [ ] `npm run build`
- [ ] implement a POST endpoint to handle the callback and gather the correct user Name and Email if required (should probably be configurable aswell, maybe you don't need the `email` and `name` scopes)
- [x] support custom fields (support clientSecret generation instead of providing it manually; it requires these fields: `privateKey`, `teamId`, `clientId` and `keyId`; the generation is implemented in `pocketbase/tools/auth/GenerateAppleClientSecret`

Fixes #202.


paulocoutinhox:
+1

n1xx1:
@ganigeorgiev 
Did you have a chance to take a loot at this? What are the issues you see with this way of handling it? Just making sure before I try to rebase to the current master.

ganigeorgiev:
@n1xx1 No, sorry I didn't check it. For now it is a very low priority. I've contacted Apple support 2-3 weeks ago to provide more information why I'm not able to enroll in the developer program, but so far there was no response from them.

In any case, you can change the upstream branch to `master`, since the `rc` will be deleted in the near future.

ganigeorgiev:
@n1xx1 At first glance I think it is ok but it could be implemented differently. I;m not sure about the introduced interface and provider abstraction (it will be also a breaking change for those who are using the Settings structs as part of a migration; I've already got enough complains after moving `core.Settings` to `settings.Settings` and I want to avoid it if possible).

If the only reason for the Apple specific fields (`teamId`, `privateKey`, etc.) is the client secret generation, then it may be better those fields to remove them from the provider and to leave the secret generation to a dedicated API endpoint (it could accept those fields as body params).

When configuring the Apple settings, the user could have a button in the Apple Options UI component that will send those fields and will generate (and maybe set in the background?) the secret, which we'll be able to store in the settings as usual without changing much the existing providers implementation.

ganigeorgiev:
I'm closing this PR since it is too much behind `master` and it will be easier to start from scratch.

I'll push a refactored version in a new PR for public test in the next couple days with the notes from https://github.com/pocketbase/pocketbase/pull/899#issuecomment-1384251733 (and maybe also `id_token` verification since Apple doesn't seem to have a UserInfo endpoint).

ganigeorgiev:
I've submitted a new PR with the refactored implementation in https://github.com/pocketbase/pocketbase/pull/1954.

I have't verified whether it works since I don't have Apple Developer account.

**@n1xx1 or anyone else who have the time and want to test it, please check the suggested steps in the PR #1954.**

PR #970, opened 2022-11-08
[Feature] Added Twitch OAuth provider

AdrianGonz97:
Hi again!

I've noticed that PRs for adding OAuth providers have been merging again and since the completion of #376, I thought I'd toss this one in again. I've updated it from my previous PR #555 to satisfy the overhaul.


ganigeorgiev:
Thank you, I'll review it and test in more details this weekend. At first glance, the only thing I'm not sure about is the custom `Twitch.FetchAuthUser` method. I'm not sure how relevant this is, but based on the comment in https://discuss.dev.twitch.tv/t/client-id-requirement-faqs/6108, it looks like it may not be required if the request has a token with it:
> If I’m including an OAuth token, do I need to include a Client-ID?
> If you are passing in an OAuth token, **we will figure out the Client-ID for you**.

If the above is not true, then we could eventually allow passing optional custom headers to the request object in the base `FetchAuthUser` method.

ganigeorgiev:
I've merged the PR locally in https://github.com/pocketbase/pocketbase/commit/4c096fd745ae1943dc3075c43a47e148e49cd69c.

To avoid breaking the Provider interface and to allow more flexibility on the request, I've decided to split the `FetchRawUserData` execution to allow passing a http.Request instance.

The provider will be available with the final v0.8.0 release (most likely at the end of the next week).

PR #974, opened 2022-11-09
[feat]: Add Kakao OAuth2 provider

teddysbeach:


tools/auth/kakao.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// FetchAuthUser returns an AuthUser instance based on the Kakao's user api.
+func (p *Kakao) FetchAuthUser(token *oauth2.Token) (*AuthUser, error) {
+	// https://developers.kakao.com/docs/latest/en/kakaologin/prerequisite#personal-information
+	rawData := struct {
+		Id         string `json:"id"`
+		Name       string `json:"kakao_account.name"`
+		Email      string `json:"kakao_account.email"`
+		ProfileImg string `json:"kakao_account.profile.thumbnail_image_url"`
```
ganigeorgiev, line 35: I haven't tested it and I'm not sure how Kakao returns the response, but if the above is intended to map nested json keys I don't think it will work. Or in other words, if Kakao response is something like:
```js
{
    "id": "...",
    "kakao_account": {
        "name": "...",
        "email": "...",
        "profile": {
            "profile_image_url": "..."
        }
    }
}
```
You'll have to create similar nested struct in Go:

```go
rawData := struct {
    Id string `json:"id"`
    KakaoAccount struct {
        Name string `json:"name"`
        Email string `json:"name"`
        Profile struct{
            ImageUrl string `json:"profile_image_url"`
        } `json:"profile"`
    } `json:"kakao_account"`
}{}
```

Also notice in the above I've replaced `thumbnail_image_url` with `profile_image_url` because the latter is slightly bigger and may be used in more use cases.

tools/auth/kakao.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Kakao allows authentication via Kakao OAuth2.
+type Kakao struct {
+	*baseProvider
+}
+
+// NewKakaoProvider creates a new Kakao provider instance with some defaults.
+func NewKakaoProvider() *Kakao {
+	return &Kakao{&baseProvider{
+		scopes:     []string{"profile_nickname", "profile_image", "account_email", "name", "gender", "age_range", "birthyear"},
```
ganigeorgiev, line 21: We need only the scopes that give us access to the mapped fields below (I guess account_email, name, and profile_image).

ganigeorgiev (changes requested):
@teddysbeach Thank you. I'll review it and test it in more details this weekend. I left some notes for when you find time.

tools/auth/kakao.go
teddysbeach: That's right. the scope specs were different from my understanding.

tools/auth/kakao.go
teddysbeach: I am not familiar with Golang yet so It took a quite long time to figure it out on the first try, but now I totally understand after seeing your code. Thanks for helping me for real.

PR #979, opened 2022-11-09
[feat]: Add Kakao OAuth2 provider *mentioned request fixed.

teddysbeach:
Thanks for taking the time to add a review for helping integrate Kakao OAuth2.
might happen to add another OAuth2 provider in near future. So it was a good chance to get closer to Pocketbase's origin.


ganigeorgiev:
@teddysbeach I've merged the changes locally in https://github.com/pocketbase/pocketbase/commit/50fce1f3cf1c82bb9c80b37a38568bc49800da32 because I had to do some minor adjustment to the tests. I've also assigned the `nickname` to the `username` field and added extra check for the email to ensure that it is Kakao verified.

The provider will be available with the final v0.8.0 release (most likely at the end of the next week).

Issue #28, opened 2022-07-08
Add CDN field to the S3 configuration


ganigeorgiev:
Add an optional CDN field to the S3 configuration so that the files api could redirect to it in order to avoid the quotas and rate limits that some S3 providers apply.

pataquets:
As a workaround, here's an idea: put some reverse proxy in front which will rewrite the returned URL to that of the CDN.
Depending on how different is the object's CDN URL from the original so it can be easily derived, you might want to give it a try.
Or maybe some middleware like min.io can handle this.
Anyway, the feature would be nice to have.

dizys:
Or a optional URL rewrite rule config, e.g.: `https://mycdn.com/{BUCKET}/{KEY}`

ganigeorgiev:
This unfortunately ended up more troublesome than I initially anticipated.

I've implemented locally a slightly modified version of @dizys's [PR#28](https://github.com/pocketbase/pocketbase/pull/986), but I've stumbled on several issues.

Redirecting to a CDN endpoint will require the objects to be public, but unfortunatelly not all S3 providers have options to control globally the visibility of an entire bucket or prefix. One workaround I found was to upload each individual blob with special write options specifying the ACL `public-read` but again, it may not be supported by every S3 provider (_it will also conflict with the planned private/protected files support in #215, but I guess we can ignore it for this specific case_).

Another bigger and blocker issue that I've stumbled on is that the uploaded files may not be always immediately available after upload via the CDN endpoint and we'll get 40x when when retrieving the file or when just sending a HEAD request to check for file existence.
I've tested it with AWS CloudFront and Digital Ocean Spaces CDN and both have the above issue. This is not isolated only for the main uploaded file, but also affects the lazy thumbs generation.

**To summarize** - I'm not sure if implemententing it as it is will be a good idea since it is not reliable enough and it is not guaranteed to work with all S3 compatible vendors.

Depending on the users demand, I may revisit it again in the future, but for now I'll close the issue and the recommended workaround to reduce the number of S3 API calls would be to enable `/api/files/*` cache on a reverse proxy level.

danibjor:
Files not being available is also an issue on local file storage. 

Scenario I’ve tested locally. Add subscription on a collection. Upload new record that contains file to this collection. When subscription event receives client, fetch record again with expanded relations, then render on screen. Most of the time image returns 404, but a 2-300ms delay before fetching the record solves it for my case.

ganigeorgiev:
[...]

PR #986, opened 2022-11-10
[Feat] Add File URL Rewriting to S3 storage

dizys:
To allow S3 or S3-compatible storage to use a CDN as downloading endpoint, this PR adds an optional `urlRewriteRule` config for S3 storage and implements a 302 redirect for the File api if the config is set.  

## Admin dashboard change screenshot
image

## Current Implementation Issue

Since the admin dashboard is using `img-src 'self' http://127.0.0.1:* data:;` as its security policy, the CDN images that are under another hostname are forbidden to be shown on the site. I currently loosen this policy up in this PR by allowing the prefix `https:`, assuming that in this case XSS for image src under https is hard to exploit. But if there's any other better way to handle this, please enlighten me.

## Related issues
#28


ganigeorgiev:
Thank you for working on this but it is a little more complicated because we'll still hit the S3 api when accessing the thumbs in the [download action - file.go#L65-L85](https://github.com/pocketbase/pocketbase/blob/90dba45d7ca53928dfd0a8c99c6aabee5239a58e/apis/file.go#L69-L85) so we'll need some other alternative way of checking whether a thumb exist (eg. send HEAD request to the CDN endpoint) and make the S3 calls to persist the new thumb only if it is missing (eg. on first thumb access).

I'll have a more detailed loo at it after the v0.8.0 release (_I'm also not sure about the input placement in the UI and whether it is clear enough, but I can't think of other options at the moment_).

dizys:
I just pushed a new commit to address the issue of stilling calling S3 api when accessing thumbs.
Here are the essential part that I added to Attributes methods: [filesystem.go#LL128-L162](https://github.com/metricsubs/pocketbase/blob/7edad2828b56bf221bcd7d21fd8c6b6564c33c79/tools/filesystem/filesystem.go#LL128-L162).
Basically, this is like what you suggested, which is to send a HEAD request to the rewriteURL and get info like "Last-Modified", "Content-Length", "Content-Type". But I'm not sure if this is enough for PocketBase's use case.

ganigeorgiev:
This will be implemented in the next release.

The approach would be similar but it needs to be structured a little differently.

ganigeorgiev:
Just for info - I've decided not to implement this for now as part of v1. You can read more why in https://github.com/pocketbase/pocketbase/issues/28#issuecomment-1492516269.

PR #999, opened 2022-11-11
record_create_api

recoilme:
Create - create single record in collection with provided fields/values (fieldsMap)
 - check rights/rules
 - don't trigger hooks/logs event
 - return error or record struct

### DIFF
--- a/apis/record_api.go
+++ b/apis/record_api.go
@@ -1,0 +1,89 @@
+package apis
+
+import (
+	"fmt"
+
+	"github.com/pocketbase/dbx"
+	"github.com/pocketbase/pocketbase/core"
+	"github.com/pocketbase/pocketbase/daos"
+	"github.com/pocketbase/pocketbase/forms"
+	"github.com/pocketbase/pocketbase/models"
+	"github.com/pocketbase/pocketbase/resolvers"
+	"github.com/pocketbase/pocketbase/tools/search"
+)
+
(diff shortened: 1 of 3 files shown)


ganigeorgiev:
I don't want to pollute the apis.* namespace with unnecessary public methods that are almost identical with the action. It will be hard to maintain both versions. The better way would be to restructure the internals in such way that they can reuse the extracted method. Could you elaborate what is your use case for this?

recoilme:
We discussed this here https://github.com/pocketbase/pocketbase/discussions/995#discussioncomment-4110673

ganigeorgiev:
I know, but https://github.com/pocketbase/pocketbase/discussions/995#discussioncomment-4110673 doesn't explain why you need such method. If you are creating a custom action why would you need to check the collection create API rule and why don't use the `OnRecordBeforeCreateRequest` hook to modify the create behavior? If you want to create record programmatically, you don't need to check for the create rule and only checking the form validations would be enough.

In any case, I don't think this would be very useful as in its current form because it's very restrictive (eg. there is no way to pass uploaded files or to check individual errors) and adds unnecessary layer of abstraction.

The only part that would be a good idea to export probably would be the `createTest` method but even then I'm not sure how this could be used.

An example for something that will be good to be exported would be the expand record helpers because currently you have to copy the same View rule check as in the record_helpers.go, but I would that this weekend before the final v0.8.0.

PR #412, opened 2022-09-11
Feat.dark theme

deleonjuan02:
### changes
1. dark-theme class added to _vars.scss.
2. new color vars for dark-theme
3. some scss files modified in order to match with dark-theme
4. option 'change theme' implemented on the user menu
5. read the theme seted on localhost in order to apply it automatically

### photos 
image
image
image

Issue #407, opened 2022-09-09
Feature Request: Dark Mode


MrDogeBro:
Just a relatively simple feature request. It would be nice if there was a UI option to enable a dark mode since the light mode can be a bit much sometimes. That's all. Just a dark mode for the eyeballs. Thanks!

ganigeorgiev:
I'll add the request to the roadmap, but it will be a low priority for now.

PocketBase is still in its very early development stage and different color scheme(s) will be hard to maintain while still doing changes in the UI. We use CSS variables, but modifying the colors usually is not enough and there will be always some components that will need special adjustment to make them visually more appealing and consistent.

innovatorved:
If you dont have any problem i would love to build the Dark mode feature

ganigeorgiev:
@innovatorved Thank you, but my advice for now would be to wait because there are expected to be a lot of changes to the Admin UI related to https://github.com/pocketbase/pocketbase/issues/376.

In general, as I mentioned, the project is still in its early development stage and it will be better to first reach a little more stable version before adding a Dark mode since it will be hard to maintain both when adding new features.

innovatorved:
> @innovatorved Thank you, but my advice for now would be to wait because there are expected to be a lot of changes to the Admin UI related to #376.
> 
> In general, as I mentioned, the project is still in its early development stage and it will be better to first reach a little more stable version before adding a Dark mode since it will be hard to maintain both versions when adding new features.

@ganigeorgiev Sure

tigawanna:
the admin dashboard works really well with the chrome dark mode flag enabled

brodriguez-eb:
Open google console and paste this code:
```js
document.children[0].style.filter = "invert(1)"
```

brewhousedigital:
> Open google console and paste this code:
> 
> ```js
> document.children[0].style.filter = "invert(1)"
> ```

I think this might be easier in the Svelte UI

```css
html {
    filter: invert(1);
}
`

**Comment by eslym** [comment]
> > Open google console and paste this code:
> > ```js
> > document.children[0].style.filter = "invert(1)"
> > ```
> 
> I think this might be easier in the Svelte UI
> 
> ```css
> html {
>     filter: invert(1);
> }
> ```

😂 add more

```css
html {
    filter: invert(1);
}
img {
    filter: invert(1); // invert img back to normal
}
```

PeterAlfonsLoch:
> Open google console and paste this code:
> 
> ```js
[...]

PR #1000, opened 2022-11-11
feat(ui): add dark theme

restray:
Hey,
I tried to add an easy-to-use and integration-free dark theme, related to #407 !

I've seen a bit late the PR #412

### New

 - Button to switch dark theme / light theme on the admin dashboard
 - The default theme is based on the system preference
 - The preference is saved in the local storage
 - A SCSS function `colorDark` to automatically convert a *light* color to a *dark* color.
 - A `theme` store that basically make a bridge between the preferences, the browser, the local storage, etc...

#### Authentification

image

image

#### Collection

image

image

#### User

image

image

#### Stats

image

#### Setup

image

In fact, it does nothing more that the PR quoted previously!

Thanks for the time to review this :)


ganigeorgiev:
As mentioned in the related issue and the other PR, I don't want to add support for dark mode at the moment because we are still making constant changes to the UI and it will be hard to maintain.

Once PocketBase reaches a little more stable version we can always refer to this and the other dark mode PR and implement it, but for now I'll close the issue.

PR #1233, opened 2022-12-11
Added health API endpoint

MarvinJWendt:
This PR adds an `/api/health` endpoint.

It responds with a `200 OK` status and adds the following body:

```
{
  "code": 200,
  "message": "API is healthy.",
  "status": "ok"
}
``` 

Everything except for the message is validated in a test. I think the message does not need to be tested, as others should not use it as a check (they should use the status code). Let me know if you see this differently, and I'll add the check.

fixes #1203

### DIFF
--- a/apis/base.go
+++ b/apis/base.go
@@ -92,7 +92,6 @@ func InitApi(app core.App) (*echo.Echo, error) {
 
 	// default routes
 	api := e.Group("/api")
-	defer bindHealthApi(app, api) // health check should always be initialized after everything else is done
 	bindSettingsApi(app, api)
 	bindAdminApi(app, api)
 	bindCollectionApi(app, api)
@@ -101,6 +100,7 @@ func InitApi(app core.App) (*echo.Echo, error) {
 	bindFileApi(app, api)
 	bindRealtimeApi(app, api)
 	bindLogsApi(app, api)
+	bindHealthApi(app, api) // health check should always be initialized after everything else is done
 
 	// trigger the custom BeforeServe hook for the created api router
 	// allowing users to further adjust its options or register new routes
--- a/apis/base.go
+++ b/apis/base.go
@@ -92,6 +92,7 @@ func InitApi(app core.App) (*echo.Echo, error) {
 
 	// default routes
 	api := e.Group("/api")
+	defer bindHealthApi(app, api) // health check should always be initialized after everything else is done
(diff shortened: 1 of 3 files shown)


apis/base.go
```diff
@@ -92,6 +92,7 @@ func InitApi(app core.App) (*echo.Echo, error) {
 
 	// default routes
 	api := e.Group("/api")
+	defer bindHealthApi(app, api) // health check should always be initialized after everything else is done
```
ganigeorgiev, line 95: Was there any particular reason why the `defer` is needed?
It should work just fine without it.

apis/health.go
```diff
@@ -0,0 +1,29 @@
...
+
+type healthApi struct {
+	app core.App
+}
+
+// healthCheck returns a 200 OK response if the server is healthy.
+func (api *healthApi) healthCheck(c echo.Context) error {
+	payload := map[string]any{
+		"code":    http.StatusOK,
+		"status":  "ok",
```
ganigeorgiev, line 25: If `status` is not required by some 3rd party tool, let's minimize the returned fields and remove it from the response since it could change in the future (the `code` and 200 response should be enough indication for healthy state).

ganigeorgiev (changes requested):
Thank you for working on this. I left some nitpicking comments.

apis/base.go
MarvinJWendt: My thought here was to delay the health check to the end, so that it's only available if everything else passed without an error, but I can change it if you want.

apis/health.go
MarvinJWendt: Done 👍

apis/base.go
ganigeorgiev: Let's remove it.

The `defer` will call the bind even if the InitApi returns an error so it shouldn't matter in this case. Additionally it is better the internal routes to be registered before the `OnBeforeServe` hook trigger so that it can return an error to the user in case they decide to register a new endpoint with the same path.

In future, we may update the healthcheck handler so that it actually checks whether the APP failed to initialize or is still in the process, but for now let's keep it simple.

MarvinJWendt: Done

ganigeorgiev:
Thank you. It is merged in `master` and it will be available with the next release (_sometime next week I'll update also the SDKs and API documentation_).

PR #1244, opened 2022-12-12
improve: optimize record references deletion

TcMits:
Hi @ganigeorgiev 

I updated to prevent fetching all referenced records. 

Hope it helps

### DIFF
--- a/daos/record.go
+++ b/daos/record.go
@@ -101,6 +101,7 @@ func (dao *Dao) FindRecordsByIds(
 // Returns an empty slice if no records are found.
 //
 // Example:
+//
 //	expr1 := dbx.HashExp{"email": "test@example.com"}
 //	expr2 := dbx.NewExp("LOWER(username) = {:username}", dbx.Params{"username": "test"})
 //	dao.FindRecordsByExpr("example", expr1, expr2)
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
@TcMits The delete handling is starting to become very "hairy".

We can add batching for the relations to reduce loading them all in the memory, but I don't fully understand the proposed changes.

Could you clarify why the timestamp check is needed? Are you using it as a offset? What would be the behavior if  a record has more than 1 referencing field?

TcMits:
>Could you clarify why the timestamp check is needed? Are you using it as an offset? What would be the behavior if a record has more than 1 referencing field?

Oh sorry, i misunderstand `deleteRefRecords` function. `dbx.HashExp{uniqueJsonEachAlias + ".value": record.Id}` is enough. I added a commit to fix it.

ganigeorgiev:
Thank you. I'll use it as a reference and implement it slightly differently to see if I can reorganize/split the whole delete handling.

I also will try to add a test since I'm starting to lose confident in the correctness of the method (we'll need to check both the batching and "page" workers processing).

Will let you know once I push a commit.

TcMits:
>Thank you. I'll use it as a reference and implement it slightly differently to see if I can reorganize/split the whole delete handling.

Great!!

ganigeorgiev:
@TcMits Just to let you know that I've pushed in master a slightly refactored version and a Dao test to check the batch concurrent processing (both when cascade delete records and when only update a field). If you stumble on something else, feel free to let me know and I'll investigate it.

TcMits:
@ganigeorgiev 

It's looked great, but why don't you using `make()` to initialize `rows`.
```go
rows := make([]dbx.NullStringMap, 0, batchSize)
```

And keep allocated memory for next batch to prevent reallocating.
```go
rows = rows[:0]
```

I have tested in my [fork](https://github.com/TcMits/pocketbase/blob/12ec8f1c38b07fe8a25413c0e25e2973dcb3211e/daos/record.go#L426)  

References:
>Here is a function to append data to a slice. If the data exceeds the capacity, the slice is reallocated. The resulting slice is returned. The function uses the fact that len and cap are legal when applied to the nil slice, and return 0.

https://go.dev/doc/effective_go#allocation_make

>To increase the capacity of a slice one must create a new, larger slice and copy the contents of the original slice into it. This technique is how dynamic array implementations from other languages work behind the scenes.

https://go.dev/blog/slices-intro

Hope it helps

ganigeorgiev:
@TcMits Because it is just more readable this way and easy to refactor in the future, but I guess we can change it.

TcMits:
> @TcMits Because it is just more readable this way and easy to refactor in the future, but I guess we can change it (at least the preallocated cap size).

@ganigeorgiev thanks, tag me if you need a PR. 😄

PR #1277, opened 2022-12-16
fix: update error string

TcMits:


ganigeorgiev:
Thank you, but I'm still not sure about the naming of `sync` and `async` and probably will change it to `concurrent` and `nonConcurrent`.
Will update the error message accordingly, depending on what I ended up choosing.

PR #1343, opened 2022-12-22
fixed s3 file upload error caused by underscore in metadata

yuxiang-gao:
The underscore in the "original_filename" metadata is causing "SignatureDoesNotMatch" error for me when uploading files to s3 storage. Changing the metadata name to "original-filename" resolved the issue.

### DIFF
--- a/tools/filesystem/filesystem.go
+++ b/tools/filesystem/filesystem.go
@@ -142,7 +142,7 @@ func (s *System) UploadFile(file *File, fileKey string) error {
 	opts := &blob.WriterOptions{
 		ContentType: mt.String(),
 		Metadata: map[string]string{
-			"original_filename": originalName,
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
That's strange. Could you provide the exact error log?

ganigeorgiev:
I was able to reproduce the issue.

I'm not sure exactly why is happening but I think it is a bug with the upstream `github.com/aws/aws-sdk-go` package. Or maybe it is just an implementation detail since internally it converts the metadata to http header key-value pairs. There is a newer version (https://github.com/aws/aws-sdk-go-v2) that we'll have to upgrade sometime in the future and I'll create an issue in the upstream library if it still persists but for now I'll merge it with the change since it could be a blocker for some users.

I'll push a v0.10.3 release sometime later today.

PR #1346, opened 2022-12-22
upgraded to aws-sdk-go-v2

yuxiang-gao:
This PR upgrades the S3 filesystem to use the newer version of AWS sdk (https://github.com/aws/aws-sdk-go-v2) since it was mentioned in #1343 that there is plan to do so.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,10 @@ go 1.18
 
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.6
-	github.com/aws/aws-sdk-go v1.44.165
+	github.com/aws/aws-sdk-go-v2 v1.17.3
+	github.com/aws/aws-sdk-go-v2/config v1.18.7
+	github.com/aws/aws-sdk-go-v2/credentials v1.13.7
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.29.6
 	github.com/disintegration/imaging v1.6.2
 	github.com/domodwyer/mailyak/v3 v3.3.4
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
I don't want to rush it since I remember reading in the docs that the new version uses newer S3 APIs which may not be supported by all S3 providers.
It will need first to be investigated whether it works with the common S3 providers and what actually are the api changes.

yuxiang-gao:
> It will need first to be investigated whether it works with the common S3 providers and what actually are the api changes.

That makes sense. I have only briefly tested it on three providers that I am using (AWS, Tencent COS, and MinIO) and they seems to be working. Let me know if you need any help investigating this.

pr0ton11:
I would be up to test it with CEPH Rados Gatwway once I get it up and running. How is performance in general in S3 buckets for Pocketbase. Is it noticable?

ganigeorgiev:
@pr0ton11 I doubt that there will be any noticable performance difference when switching from v1 to v2 of the s3 library.

If you are asking for general S3 integration performance - it depends on the internal network infrastructure of your host provider(s), the location/region of the servers, etc.

It will be definitely slower compared to a local filesystem storage.

HiteshGorana (approved):

ganigeorgiev:
I've squash merged it in the `develop` branch with one minor endpoint normalization (prefixing it with `https://` if scheme is not defined; it seems that this was performed behind the scenes by v1).

I've tested it with AWS S3, DigitalOcean Spaces, Vultr Object Storage and Wasabi S3 offerings and all seems to work fine.

PR #1370, opened 2022-12-24
Fix textarea enter keydown

tenthree:
When a user types CJK characters and the IME show up, pressing ENTER to select a word will also trigger form submission(save changes) simultaneously; to avoid unexpected request submission, we could add the `event.isComposing` checking to resolve it, thx.

[MDN references](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing)

pb


ganigeorgiev:
Thank you. I wasn't aware of IME and it specific events.

Could you pull the latest changes from master and rebuild the ui/dist?

ganigeorgiev:
I've squash merged it. The fix will be available with the next bugfix or v0.11.0 release (sometime next week or the week following it).

tenthree:
> I've squash merged it. The fix will be available with the next bugfix or v0.11.0 release (sometime next week or the week following it).

No problem, thx.

PR #1375, opened 2022-12-25
make it so the avatar moves with the cursor

Merkie:


Merkie:
so basically i just added some javascript that makes it so the avatar in the corner moves with the mouse cursor

ganigeorgiev:

PR #1377, opened 2022-12-25
Add authentik oauth provider

pr0ton11:
This is a PR that eventually should add a more generic oauth2 auth provider. Since I am self hosting [authentik](https://goauthentik.io/) as my identity provider it only makes sense that I implement auth for it in projects I want to use in the future.

- [x]  Implement FetchAuthUser functionality
- [x] Add unit tests
- [x] Improve UI to be more aligned with vision of pocketbase team (help needed)


tools/auth/authentik.go
```diff
@@ -0,0 +1,34 @@
...
+// Name of the oauth provider
+const NameAuthentik string = "authentik"
+
+type Authentik struct {
+	*baseProvider
+}
+
+func NewAuthentikProvider() *Authentik {
+	return &Authentik{&baseProvider{
+		scopes:   []string{"profile"},
```
mariosant, line 25: Correct me if I am wrong please, but I feel that scopes can be preconfigured on provider side. If that's the case, it might make sense to not request specific scopes, as it will override the provider's configuration (while pocketbase request scopes are not configurable atm).

Ref: https://goauthentik.io/docs/providers/oauth2/#scope-authorization

pr0ton11: You are partially correct. This default implementation just requests a scope that is enabled by default in Authentik, but ofc this could be configured in any way on the provider side.
I am struggling at the moment to find where the frontend settings are passed into this class. If someone could give me a hint that would be very helpful for the progress of this PR.

ganigeorgiev: > I am struggling at the moment to find where the frontend settings are passed into this class. If someone could give me a hint that would be very helpful for the progress of this PR.

I'm not sure if that's what you are asking, but users can't configure the default `scopes` from the Admin UI at the moment. At the end, they are just minimal defaults for convenience and are used only in the login url(s) that are returned with the [`List auth method` endpoint](https://pocketbase.io/docs/api-records/). Users can adjust/replace them client-side by modifing the `scope` query parameter in the provider login url(s).

pr0ton11: Thanks @ganigeorgiev 

- What I meant is that Authenktik.go uses a *baseProvider instance, but I dont see where in the baseProvider instance I can retrieve the settings that get set in the Svelte frontend component AuthentikOptions.svelte. This is essential, since all the endpoints get configured there (and authentik is self-hosted, so I can not provide a general URL in the constructor)
- Maybe you could also give me a hint since authentik is a very general provider, and could also allow for example Keycloak to be used, if we want to keep it named Authentik or if we want to give it a more general name.
- I need some feedback about the frontend part

ganigeorgiev: @pr0ton11 The Admin UI settings API doesn't operate directly with the `auth.baseProvider` and instead uses the `settings.AuthProviderConfig` - https://github.com/pocketbase/pocketbase/blob/c1921aeef8b28f9df66d5be3143e0c39dd079bec/models/settings/settings.go#L461-L468

You can find details about the available settings API fields in https://pocketbase.io/docs/api-settings/.

pr0ton11: Thanks for helping with the PR, I completed the implementation and it looks like everything is working as expected now

ganigeorgiev:
@pr0ton11 Thank you for working on this but there are some minor things that need to be improved:

1. Add the authentik in the `apis/settings_test.go` tests.
2. Please copy the comments format in authentik.go to be godoc "compatible" (you can refer to the other provider as a reference).
3. Map the `email_verified` field and assign the extracted email only if it is true.
4. **There is no `id` field in the authentik response**, and you'll always get `"0"` at least by default. There is a `sub` field but I'm not sure if this is user unique since I couldn't find anywhere in the authentik docs. I have no experience with authentik so please verify and confirm. We need a unique user identifier so that we can link the authentik user with the PocketBase one.

ganigeorgiev:
@pr0ton11 I'll apply the changes manually to skip the back-and-forth and will merge it in the develop branch sometime later today.

ganigeorgiev:
@pr0ton11 I've squash merged it in the develop branch - https://github.com/pocketbase/pocketbase/commit/6d08a5f36fb92ac00409c9405a7c8d0d5f9f18d5.

It will be included in the next v0.12.0 release.

pr0ton11:
Thanks for your effort for the final changes 👍

cgrs:
FYI, I've tested this OAuth2 provider with [Zitadel](https://zitadel.com/) and [Keycloak](https://www.keycloak.org/) and they worked successfully. I think this provider should be referred as “Generic” as long as more OAuth2 providers can plug in (I'd like to test it also with Ory, but it should work out of the box).

pr0ton11:
> FYI, I've tested this OAuth2 provider with [Zitadel](https://zitadel.com/) and [Keycloak](https://www.keycloak.org/) and they worked successfully. I think this provider should be referred as “Generic” as long as more OAuth2 providers can plug in (I'd like to test it also with Ory, but it should work out of the box).

Maybe support for the .well-known URI could be added and it could be renamed to generic.
At least the documentation should mention this as a more generic oauth2 provider

ganigeorgiev:
@cgrs, @pr0ton11 Let's keep it for now as it is.

Unfortunately, there will be always some manual mapping required when fetching the user profile data that will defer from one provider to another, and I'm not sure if we can have a truly generic provider.

_Eventually, I'll consider adding a note/label in the UI mentioning that the provider could be used with other similar OAuth2 services._

ganigeorgiev:
@cgrs, @pr0ton11 On second thought having a generic OIDC provider may actually be better since the OIDC spec mentions recommended/standard scope claims and fields - https://openid.net/specs/openid-connect-core-1_0.html

After the ongoing "View" collection type, I'll consider generalizing the Authentik provider.

PR #1380, opened 2022-12-25
Feature: Add dimensions validation for images (width x height)

### LINKED ISSUE #1266: Feature request: Limit image width and height
Instead of only limiting file size, I would like to be able to require the uploaded image to be restricted by width and height rules.

I have a preview component, that expects an image of specific size, so I would like to prevent user from uploading image that is not of that size. But preventing images that are smaller than certain resolution would also be helpfull.

AmFlint:
## Description

In this PR, I added a new Schema field to files (when creating/updating a collection), which allows users to specify "accepted" dimensions for images (width x height) uploaded by users.

When a user "upserts" a record, if this list of accepted dimensions is configured on the collection, then PocketBase checks the file's width/height to make sure these dimensions match at least one of the "accepted dimensions". (see screenshots)

This new validation only works on "image" files, it will fail if the uploaded file is not an image (the code uses the image standard library to get the dimensions of the image for the validation logic).

Also, I refactored test's "Mock" function to either create a text file (used in many tests) or an image file (added to the repository) with the dimensions 100x100. This allows me to test my validation logic, and will allow developers to test image uploads in the future.

## Screenshots

UI when creating/updating a collection:
Screenshot 2022-12-25 at 13 29 57

UI when validation failed when uploading an image that does not match the accepted dimensions set on the collection field:
Screenshot 2022-12-25 at 13 30 27

## Issues

- Resolves https://github.com/pocketbase/pocketbase/issues/1266


ganigeorgiev:
Hi, I appreciate the effort, but please refrain creating PR for new features without previously discussing the implementation details (also please refer to https://github.com/pocketbase/pocketbase#contributing).

Currently https://github.com/pocketbase/pocketbase/issues/1266 is not planned and is not in the [roadmap](https://github.com/orgs/pocketbase/projects/2) because I'm still not sure whether such validator is a good idea and how it will work.

At the moment it seems that if this validator is specified and the user tries to upload a file, it will always fail (because the decode will fail), even though the file is allowed in the mimetypes list.

We need first to decide:

- do we really need such validator and is it clear being part of the `file` type field
- do we want a fixed size validator or min/max width/height

I'll close the issue for now to prevent triggering notifications in case of new changes in master, but we can refer to it later in case we decide to include it.

PR #1420, opened 2022-12-29
Add filesystem.NewFileFromBytes

dschissler:
This PR adds tools/filesystem.NewFileFromBytes which should greatly help uploading files from events.  There are a lot of libraries which output bytes.  It's pointless to save it to a file just to read it again.

I was also not able to test this, nor was I able to even compile it.  Baby steps.

### DIFF
--- a/tools/filesystem/file.go
+++ b/tools/filesystem/file.go
@@ -1,6 +1,7 @@
 package filesystem
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"mime/multipart"
@@ -46,6 +47,18 @@ func NewFileFromPath(path string) (*File, error) {
 	return f, nil
 }
 
+// NewFileFromBytes creates a new File instance from the provided byte slice.
+func NewFileFromBytes(b []byte, name string) (*File, error) {
+	f := &File{}
+
+	f.Reader = &BufReader{b}
+	f.Size = int64(len(b))
+	f.OriginalName = name
+	f.Name = normalizeName(f.Reader, f.OriginalName)
+
+	return f, nil
+}
+
 // NewFileFromMultipart creates a new File instace from the provided multipart header.
 func NewFileFromMultipart(mh *multipart.FileHeader) (*File, error) {
 	f := &File{}
@@ -83,6 +96,31 @@ type PathReader struct {
 // Open implements the [filesystem.FileReader] interface.
 func (r *PathReader) Open() (io.ReadSeekCloser, error) {
 	return os.Open(r.Path)
+}
+
+// -------------------------------------------------------------------
+
+var _ FileReader = (*BufReader)(nil)
+
+type BufReader struct {
+	Bytes []byte
+}
+
+// Open implements the [filesystem.FileReader] interface.
+func (r *BufReader) Open() (io.ReadSeekCloser, error) {
+	return NewBytesReadSeekCloser(r.Bytes), nil
+}
+
+type BytesReadSeekCloser struct {
+	*bytes.Reader
+}
+
+func NewBytesReadSeekCloser(b []byte) *BytesReadSeekCloser {
+	return &BytesReadSeekCloser{bytes.NewReader(b)}
+}
+
+func (r *BytesReadSeekCloser) Close() error {
+	return nil
 }
 
 // -------------------------------------------------------------------
--- a/tools/filesystem/file.go
+++ b/tools/filesystem/file.go
@@ -51,7 +51,7 @@ func NewFileFromPath(path string) (*File, error) {
 func NewFileFromBytes(b []byte, name string) (*File, error) {
 	f := &File{}
 
-	f.Reader = &BufReader{b}
+	f.Reader = &bytesReader{b}
 	f.Size = int64(len(b))
 	f.OriginalName = name
 	f.Name = normalizeName(f.Reader, f.OriginalName)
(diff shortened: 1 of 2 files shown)


tools/filesystem/file.go
```diff
@@ -87,6 +100,31 @@ func (r *PathReader) Open() (io.ReadSeekCloser, error) {
 
 // -------------------------------------------------------------------
 
+var _ FileReader = (*BufReader)(nil)
+
+type BufReader struct {
```
ganigeorgiev, line 105: For consistency with the `New*` method let's change this to `BytesReader`.

tools/filesystem/file.go
```diff
@@ -87,6 +100,31 @@ func (r *PathReader) Open() (io.ReadSeekCloser, error) {
...
+// Open implements the [filesystem.FileReader] interface.
+func (r *BufReader) Open() (io.ReadSeekCloser, error) {
+	return NewBytesReadSeekCloser(r.Bytes), nil
+}
+
+type BytesReadSeekCloser struct {
+	*bytes.Reader
+}
+
+func NewBytesReadSeekCloser(b []byte) *BytesReadSeekCloser {
```
ganigeorgiev, line 118: Since this is currently used only inside the `Open()`, let's change it to `newBytesReadSeekCloser()` to limit the surface of the exported package members (or remove it and directly call the returned struct literal in its place).

tools/filesystem/file.go
```diff
@@ -87,6 +100,31 @@ func (r *PathReader) Open() (io.ReadSeekCloser, error) {
...
+	return &BytesReadSeekCloser{bytes.NewReader(b)}
+}
+
+func (r *BytesReadSeekCloser) Close() error {
```
ganigeorgiev, line 122: Add a comment that this "implements" the `io.ReadSeekCloser` interface to be a little more clear why this method is here, eg:
```
Close implements the [io.ReadSeekCloser] interface.
```

_I'm still not sure whether it would be better reset/rewind the seeker, but let's keep it empty for now._

ganigeorgiev (changes requested):
I left some nitpicking comments.

For the tests - similarly to the other factories, we need a test at least for the `NewFileFromBytes` (you can check `file_test.go`).

To run the tests, you can execute something like:
```sh
# run all filesystem tests
go test ./tools/filesystem/...
```

---

To test it manually with your custom code, you can use `replace` in your go.mod, eg:

```go.mod
// go.mod
module yourapp

go 1.18

replace github.com/pocketbase/pocketbase => ../forked_pocketbase

require(
...

```

dschissler:
@ganigeorgiev Have another look.

tools/filesystem/file.go
```diff
@@ -46,6 +48,23 @@ func NewFileFromPath(path string) (*File, error) {
 	return f, nil
 }
 
+// NewFileFromBytes creates a new File instance from the provided byte slice.
+func NewFileFromBytes(b []byte, name string) (*File, error) {
+	size := len(b)
+	if size == 0 || b == nil {
```
ganigeorgiev, line 54: There is no need to check `b == nil`. `size == 0` is enough.

tools/filesystem/file.go
```diff
@@ -87,6 +106,27 @@ func (r *PathReader) Open() (io.ReadSeekCloser, error) {
 
 // -------------------------------------------------------------------
 
+var _ FileReader = (*bytesReader)(nil)
+
+type bytesReader struct {
```
ganigeorgiev, line 111: Similar to `PathReader` and `MultipartReader` reader, let's export this because we may use it as embedded struct or as a return type in some other code.

The `bytesReadSeekCloser` can remain unexported.

tools/filesystem/file_test.go
```diff
@@ -43,6 +43,33 @@ func TestNewFileFromFromPath(t *testing.T) {
 	}
 }
 
+func TestNewFileFromBytes(t *testing.T) {
+	testDir := createTestDir(t)
+	defer os.RemoveAll(testDir)
```
ganigeorgiev, line 48: I think we can remove this, since we are not using the testDir in the test.

tools/filesystem/file_test.go
```diff
@@ -43,6 +43,33 @@ func TestNewFileFromFromPath(t *testing.T) {
...
+		t.Fatal("Expected error, got nil")
+	}
+
+	// zero bytes
+	if _, err := filesystem.NewFileFromBytes([]byte{}, "photo.jpg"); err == nil {
+		t.Fatal("Expected error, got nil")
+	}
+
+	originalName := "notes.txt"
+	f, err := filesystem.NewFileFromBytes([]byte("TODO: Write Tests\n"), originalName)
```
ganigeorgiev, line 61: Can we change the text to just `test` or something else shorter?
_(Additionally, I also have a regex pattern for plain-tasks that extracts "todo" strings from the code and will get triggered unnecessary)_

tools/filesystem/file_test.go
```diff
@@ -43,6 +43,33 @@ func TestNewFileFromFromPath(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if f.Size != 18 {
+		t.Fatalf("Expected Size %v, got %v", 18, f.Size)
+	}
+	if f.OriginalName != originalName {
+		t.Fatalf("Expected originalName %q, got %q", originalName, f.OriginalName)
+	}
```
ganigeorgiev, line 70: Can you also add a test for the "normalized" name?
You can use "originalName" with some strange characters and then check it against a regex pattern (we do something similar in the existing tests).

ganigeorgiev (changes requested):
Thank you for adding the test. I left some notes.

dschissler:
@ganigeorgiev I updated it and it works in my project.  Can you merge this now?

Issue #215, opened 2022-07-25
Feature request: Private/Protected Files


bnert:
As the title suggests, are private/protected files currently scoped for 1.0.0 release? I was playing around with the files API, and as of right now, it seems like once files are uploaded they are considered public, unless I have missed something in the docs.

If this feature set is not implemented, are you open to a contribution for this feature?

As an aside, thank you so much for this project. It is AWESOME! 😃

ganigeorgiev:
That's correct - all files are considered public if the user has access to its url address. This recently discussed in - https://github.com/pocketbase/pocketbase/discussions/210.

Yes, I'm planning to add support for password protected files. I'll use this issue and will add it to the roadmap, but for now it is a low priority.

Contributions are welcomed, but I'm not sure about the implementation details at the moment, so please feel free to share your idea before proceeding.

What I had in mind was to have a new endpoint that users could request  to generate a **short-lived/one-time tokens** and pass this token as a query parameter to the file url where in the file serving endpoint we'll have to check if the related file field is protected or not and whether the user has access to it. The access permissions could be defined as an optional input (similar to the collection API rules).

newbeelearn:
If we had cookies for sesssion instead of token than this would become simpler to solve?

ganigeorgiev:
@newbeelearn It depends from a lot of things.

In order to pass the authorization token as cookie securely we will need to set at least the `Secure` and `SameSite=Strict` attributes because the other alternatives would require to implement some sort of CSRF protection, which'll further complicate sending requests to the API. And even if we use "SameSite=Strict" cookie then another problem arise and that's it from where the request originated - was it from the same domain? was it requested from a mobile app?, etc.

So while cookies could solve part of the problem they will also open a whole bunch of other issues that we'll have to take care.

bnert:
Agreed, I think a token is more apt. Nonetheless, a great question @newbeelearn!

I am going to think a bit more about implementation details and poke around the code to get a better idea of how this may be accomplished.

ganigeorgiev:
@bnert You could get a general idea how to generate the token and to restrict the file download action, by following this example - https://github.com/pocketbase/pocketbase/discussions/254#discussioncomment-3300330

The example is using the event hooks, but ideally something like this will be integrated in PocketBase itself with the main difference that instead of checking a single specific profile field, we would check a file API rule but the general idea is the same.

bnert:
[...]

PR #1437, opened 2022-12-30
Check browser cache before process collection resources

hungcrush:
I realize that before api download return data, it has to get collection data -> check thumbnail (via S3 too slow).
I just want to make full use of browser caching.. please give me your opinion 😬

### DIFF
--- a/apis/file.go
+++ b/apis/file.go
@@ -29,6 +29,14 @@ func (api *fileApi) download(c echo.Context) error {
 		return NewNotFoundError("", nil)
 	}
 
+	// check browser cache and return 304 if possible
+	ifModifiedSince := c.Request().Header.Get("If-Modified-Since")
+	if ifModifiedSince != "" {
+		c.Response().Header().Set("Last-Modified", ifModifiedSince)
+		c.Response().WriteHeader(304)
+		return nil
+	}
+
 	recordId := c.PathParam("recordId")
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
I agree that we can improve the S3 flow but I'm not sure about the suggested change and I'll have to research it in more details.

As it is in the PR right now, it seems that it will always return 304 no matter whether the file exist or whether it was actually cached, as long as the client send a `If-Modified-Since` header.

Additionally, things may get complicated when we introduce the private/protected files from #215 in the near future.

I'll have a more detailed look at this when I start working on #215 and the other S3 related requests like #28 (https://github.com/pocketbase/pocketbase/pull/986)

ganigeorgiev:
After some local tests, I don't think the eager 304 response is needed because all files are served with default 30days Cache-Control header and as far as I can see from inspecting the network requests,`If-Modified-Since` is sent only for revalidations, after the cache has expired (aka. ideally the browser shouldn't make any cached file requests to the server).

In any case, if you want always to return 304 on existing `If-Modified-Since` request header, you can define a global middleware and handle the response from there (_but as mentioned previously, I don't think this is correct in general because this way you are not making sure that the file exist or has changed based on the provided timestamp_). You can find an example how to create a middleware in https://pocketbase.io/docs/router/#middlewares.

PR #1439, opened 2022-12-30
perf: minor optimizations

TcMits:


ganigeorgiev:
Thank you. I'll apply the map size hinting in the https://github.com/pocketbase/pocketbase/tree/filter-enhancements branch, but I don't think the buffered channel is needed.

PR #1443, opened 2022-12-31
Implement strava oauth2 as new auth provider

szsascha:


ganigeorgiev:
Thank you. I've merged it locally in `master` and it will be available with the next release (_most likely sometime next week_).

PR #1448, opened 2022-12-31
Added gitee OAuth Provider

yuxiang-gao:
Happy New Year!

This PR implements OAuth provider for [Gitee](https://gitee.com/language/en), a Chinese code hosting platform similar to Github. It is fairly popular among Chinese developers and companies. This would allow me to use Pocketbase in my applications targeting a more regional audience.

Their authentication flow is fairly similar to Github's. However, their [documentation](https://gitee.com/api/v5/oauth_doc#/) seems to be only in Chinese. So if you need any help testing this I would be happy to help.


ganigeorgiev:
Thank you for working on this.

I've tested it locally and it works fine with one minor detail - when the user checks the "Keep my email address private" checkbox from the Gitee Emails management the email field is not an empty string but the text "未公开邮箱" (google translate: "undisclosed email") and this prevents making the additional emails request (and eventually fail the validation for new OAuth2 users).

I suggest:

- Remove the `Email` field from the `extracted` struct.
- Always make a request for `https://gitee.com/api/v5/emails`, bit in case of a failure, don't return an error and silently ignore it (if possible check for scope error) and just return the user as it is. This is because Gitee allows the user to uncheck the "all emails" checkbox when faced with the Auth screen. This is OK with PocketBase because we no longer require the email and the OAuth2 email is used primarily for existing accounts linking (and when the auth collection has the "Always require email" toggle, but that will trigger the validations anyway). _I'll apply a similar change to the GitHub provider sometime later_.

ganigeorgiev:
Alternatively (and probably easier) would be to just remove the extra email call (if the user prefer their email to be private then let it be).

Or in other words, keep the `Email` field in the `extracted` struct but also add a validation for it in case it is not an empty string:

```go
user := &AuthUser{
	Id:          strconv.Itoa(extracted.Id),
	Name:        extracted.Name,
	Username:    extracted.Login,
	AvatarUrl:   extracted.AvatarUrl,
	RawUser:     rawUser,
	AccessToken: token.AccessToken,
}

if extracted.Email != "" && is.EmailFormat.Validate(extracted.Email) == nil {
	user.Email = extracted.Email
}
```

ganigeorgiev:
3rd option would be to combine the above 2 suggestions, aka. make the extra api call request if the extracted email is invalid (the request must be **optional**, aka. ignore the response error and directly return the user without an email).

yuxiang-gao:
Hi, I have pushed a fix. I have chosen the 3rd option you proposed with validation and an optional request step from the emails api. All errors in the optional request step are ignored.

ganigeorgiev:
> All errors in the optional request step are ignored.

Only the optional request error should be ignored. I'm working on it locally since I'm updating also the GitHub provider.

yuxiang-gao:
Ok, I can sync your update to the Github provider to the Gitee provider once you have finished. But I think it is reasonable to ignore the subsequent io and JSON parsing error following the optional request. An other option is to validate the status code?

yuxiang-gao:
Ok, I see that you already added the status code validation.

yuxiang-gao:
I have pushed a new change to conform with the way the Github provider handles this.

ganigeorgiev:
@yuxiang-gao Please be patient, I'm currently working on it. Gitee is not the same as GitHub. It seems to return 401 in case the "Access your email data" was not checked.

Additionally, when "Keep my email address private" is set, the optional call seems that will always fail the json parsing because the returned email and response code is something like:

```js
{"email":"12285415+test@user.noreply.gitee.com"} 200
```

Which is not what we want. I'm trying to figure out all edge cases and will push and update once I test it properly.

ganigeorgiev:
I've merged it locally in `master` - https://github.com/pocketbase/pocketbase/commit/c673d9d314e1b5556e7c24392aef5c400e382694

It seems that Gitee users have 3 options (some of which can be combined):
A. "Access your emails data" (Auth screen's checkbox)
B. "Public email address" (select in the account "Emails management" page)
C. "Keep my email address private" (checkbox in the account "Emails management" page; overwrites 2)

The email is returned in the main response only if 2) is set.
The second optional request works only if  "Keep my email address private" is not set.

| A (auth screen) | B ("Public email address") | C ("Keep my email address private") | email |
|:----------------|:---------------------------|:------------------------------------|:------|
| `true`          | `true`                     | `false`                             | ✅     |
| `false`         | `true`                     | `false`                             | ✅     |
| `false`         | `false`                    | `false`                             | ❌     |
| `true`          | `false`                    | `false`                             | ✅     |
| `false`         | `false`                    | `true`                              | ❌     |
| `true`          | `false`                    | `true`                              | ❌     |

PR #1479, opened 2023-01-03
Fix typos in `ListApiDocs.svelte`

simonwiles:


ganigeorgiev:
Thanks. I'll apply the change locally in the upcoming v0.11.0 branch (`filter-enhancements`) since it will require rebuilding the ui/dist files.

Issue #976, opened 2022-11-09
Improve relations picker UI


ganigeorgiev:
##### Discussed in https://github.com/pocketbase/pocketbase/discussions/975

This could be eventually combined with https://github.com/pocketbase/pocketbase/issues/881 and https://github.com/pocketbase/pocketbase/issues/100

In general, we can try to improve the relations picker UI and eventually allow inline record preview/edit.

We can also try to add support for drag&drop sorting in case of a multi-relation field.

---

<div type='discussions-op-text'>

<sup>Originally posted by **bloor** November  9, 2022</sup>
A current issue I`m facing is , when I access a record I don;t have any direct-access to the relationed-records.
A link or "Ctrl-click" feature on it to open the linked record seems to be very helpful.

Now I have to copy the ID and manually search it in the relationed table.

![image](https://user-images.githubusercontent.com/1420653/200853136-19631a3e-631d-412d-b5f3-5c81c2cdcb45.png)
</div>

cyrilghali:
I had the same problem at first and discovered the i icon, if you hover your mouse over it you will see the json of the object!

hungcrush:
I have an idea. each task in roadmap can you write some idea, suggest, .. what will you do to solve that. Then we can make Pull Request then you will review for us 😬 
For example color Picker, Address picker. I think I can do it

ganigeorgiev:
@hungcrush Please refer to https://github.com/pocketbase/pocketbase#contributing.

PRs for new features are not always helpful, especially when they require UI changes since often I don't have a clear idea what and how things should look like and usually I end up experimenting with different approaches before choosing the one I think would fit best.

PocketBase is a hobby project. I don't want to maintain a team or spending days reviewing PRs. As mentioned in the note in the readme, I work on issues in specific order and often such PRs comes out of nowhere and skew all initial planning since they require focus change.

The relation picker UI changes are planned for v0.12/v0.13 but there are no fixed ETAs. The upcoming next release (v0.11) is focused mainly on filter enhancements (_unfortunately I wasted last week with the multi-match implementation which ended up very unefficient I'm currently rewriting it, so it will take a way before I finalize it and unit test it properly_).

ganigeorgiev:
The relations UI was refactored in the `develop` branch and it will be available with the next v0.12.0 release soon.

PR #1507, opened 2023-01-06
Add relation field name to list view

sadovyiov:
Changed ID into the relation column to a readable view

CleanShot 2023-01-06 at 17 07 19@2x


ganigeorgiev:
Thank you, but once again, I had something slightly different in mind.

I'm planning to handle it together with the more generic relations picker refactoring in v0.12/v0.13 as part of https://github.com/pocketbase/pocketbase/issues/976.

_Please file an issue/discussion with the planned implementation details before starting work on a PR for a new feature to avoid wasting your time. Also please refer to https://github.com/pocketbase/pocketbase#contributing._

PR #1513, opened 2023-01-06
Fixed SQL syntax error in FindCollectionReferences

zzaba:
Minimal reproducible example:
```
package main

import (
	"fmt"
	"os"

	"github.com/pocketbase/dbx"
	"github.com/pocketbase/pocketbase/tools/list"
)

func main() {
	app := pocketbase.New()

	app.OnAfterBootstrap().Add(func(e *core.BootstrapEvent) error {
		excludeIds := []string{""}
		dao := app.Dao()
		query := dao.CollectionQuery()
		if len(excludeIds) > 0 {
			uniqueExcludeIds := list.NonzeroUniques(excludeIds)
			// if len(uniqueExcludeIds) > 0 {
				query.AndWhere(dbx.NotIn("id", list.ToInterfaceSlice(uniqueExcludeIds)...))
			// }
		}
		fmt.Fprintln(os.Stderr, query.Build().SQL())
		return nil
	})

	app.Start()
}
```
Running this with `go run main.go >/dev/null` generates erroneous SQL:
```
SELECT {{_collections}}.* FROM `_collections` WHERE ()
```
Whereas if we uncomment the if statement generated SQL becomes correct:
```
SELECT {{_collections}}.* FROM `_collections`
```

This minor issue causes cryptic `near ")": syntax error` error if we for example try to delete collection with an empty Id field like this:
```
dao.DeleteCollection(&models.Collection{})
```

### DIFF
--- a/daos/collection.go
+++ b/daos/collection.go
@@ -87,7 +87,9 @@ func (dao *Dao) FindCollectionReferences(collection *models.Collection, excludeI
 	query := dao.CollectionQuery()
 	if len(excludeIds) > 0 {
 		uniqueExcludeIds := list.NonzeroUniques(excludeIds)
-		query.AndWhere(dbx.NotIn("id", list.ToInterfaceSlice(uniqueExcludeIds)...))
+		if len(uniqueExcludeIds) > 0 {
+			query.AndWhere(dbx.NotIn("id", list.ToInterfaceSlice(uniqueExcludeIds)...))
+		}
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
I'll have to think a little more on this.

I'll consider updating it slightly differently in the near future since there are also other places where we can add similar check for the nonzerounique results + we need unit tests.

PR #1548, opened 2023-01-09
PDF Preview

mjadobson:
As discussed in #1542

Adapts the image preview panel code to support PDF embedding using native <object> embed.
Also allows click opening of non-previewable files from collection rows

Due to current CSP rules, it will only show the embed if running the UI on the same host as PB.


ui/src/components/base/PreviewPopup.svelte
```diff
@@ -19,14 +21,20 @@
...
         <button type="button" class="overlay-close" on:click|preventDefault={hide}>
             <i class="ri-close-line" />
         </button>
     </svelte:fragment>
 
-    <img src={url} alt="Preview {url}" />
+    {#if type === "image"}
+        <img src={url} alt="Preview {url}" />
+    {:else if type === "pdf"}
+        <object title={url.substring(url.lastIndexOf("/") + 1)} data={url} type="application/pdf">
```
ganigeorgiev, line 34: Since `url.substring(url.lastIndexOf("/") + 1)` is used below too, I think we can move it as computed/reactive declaration, something like:

```js
$: filename = url ? url.substring(url.lastIndexOf("/") + 1) : "";
```

ui/src/components/records/RecordFilePreview.svelte
```diff
@@ -8,35 +8,35 @@
...
-
-    $: hasPreview = CommonHelper.hasImageExtension(filename);
-
-    $: if (hasPreview) {
-        originalUrl = ApiClient.getFileUrl(record, `${filename}`);
-    }
+    let originalUrl = ApiClient.getFileUrl(record, `${filename}`);
 
+    $: type = CommonHelper.getFileType(filename);
+    $: hasPreview = /^(image|pdf)$/.test(type);
```
ganigeorgiev, line 14: I don't think the regex is needed in this case. It would be more readable with direct checks or something like `["image", "pdf"].includes(type)`.

ui/src/components/records/RecordFilePreview.svelte
```diff
@@ -8,35 +8,35 @@
...
-        }}
-        on:error={onError}
-    />
-{:else}
-    <i class="ri-file-line" />
-{/if}
-
-<PreviewPopup bind:this={previewPopup} />
+<a
+    class="thumb thumb-sm link-fade"
```
ganigeorgiev, line 23: The classes will interfere in other places like the preview in the Record create/update panel where we use a different thumb size so we need to make it configurable if we want to keep it this way.

One way could be to define a `export let size = ''; // sm/lg` prop and change the tag to:

```html
<a
    class="thumb thumb-{size} link-fade"
```

ui/src/components/records/RecordFieldCell.svelte
```diff
@@ -57,9 +57,7 @@
     {:else if field.type === "file"}
         <div class="inline-flex">
             {#each CommonHelper.toArray(record[field.name]) as filename, i (i + filename)}
-                <figure class="thumb thumb-sm">
-                    <RecordFilePreview {record} {filename} />
-                </figure>
+                <RecordFilePreview {record} {filename} />
```
ganigeorgiev, line 60: Since now this will be always rendering a thumb, I think we can rename it to `RecordFileThumb` to be more clear.

We also will need to remove the `figure` wrapping inside the RecordUpsert panel > FileField (_you can skip the `class:fade` check_).

ganigeorgiev (changes requested):
I've tested it briefly and it is ok but it needs some minor adjustments since it currently results in double wrapping inside the RecordUpsert panel.

_I also have some concerns about the security in case someone decide to name a malicious svg/html file as `.pdf` and I'm not sure how the `object` tag will render it but I'll test it sometime later today in more details._

mjadobson:
I've made the suggested amendments. Thanks for your prompt feedback.

Having just looked into it, I'm also not sure about the security implications of using an `<object>` embed.

Perhaps `<iframe src=".." sandbox>` is a better idea. It has good support: https://caniuse.com/iframe-sandbox .

ganigeorgiev:
I've tested it and the use of the `object` tag in our case is safe because we've taken 2 security measures:

1. The `object` tag seems to preview only files that are served with `Content-Disposition: inline` and we serve `inline` only a restricted set of mimetypes (detected from the filecontent and not just from the extension):
https://github.com/pocketbase/pocketbase/blob/ede7804a80a37649f82b32eab53d693e67406ffd/tools/filesystem/filesystem.go#L270-L280
In this case, if a malicious user upload a svg/html/etc. with `.pdf` extension, it will not have a preview and will either result in a blank screen + download (in Chrome) or just the fallback text.

2. Even if in the future browsers decide to allow serving `attachment` content with the `object` tag, we always have a default CSP with the file response that prevents scripts execution: https://github.com/pocketbase/pocketbase/blob/ede7804a80a37649f82b32eab53d693e67406ffd/tools/filesystem/filesystem.go#L317

_I'll merge it as it is in the `develop` because I want to do some further minor adjustments to allow also audio and video files in the preview since we already use the `object` tag and may as well preview them._

Issue #898, opened 2022-11-02
Feature Request: Add-ons in UI


kennethklee:
An add-ons section in the admin UI would be nice.

- The Core API could have an add-on register/define function that takes in add-on details and path to web-component. or perhaps scan an add-ons directory.
- UI would then populate side menu with registered add-ons, which would navigate to the web-component page.
- web-component page is free do whatever, perhaps use its own pocketbase sdk with `pb-admin-auth` storage key.

I'd love to help with this after v0.8.0.

ganigeorgiev:
It is in my todo but until there is a clear definition what and how things will work it remains out of the scope for v1.0.0.

Having just a plain "placeholder" where to put a custom html is not enough.

This may looks easy at first glance but it's not. We'll need to:
(including the styles)
- be able to provide access to the admin SPA router and some of the app svelte stores
- ideally, convert the existing admin svelte components in their own self-contained package and/or individual web-component so that they can be reused and provide a more unified UX (eg. alerts, popups, errors, etc.)

Matrix89:
Hi, is there any way I can help with this?

drudge:
I would love this.

Anatoly03:
(Sorry if the below sounds like an order, I'm used to describing features in a "It should" manner. Consider it more like a detailed idea of how such a thing could be implemented.)

This could be implemented by having a directory of extensions which would be Node.JS packages or submodules to such. An extension could have some kind of `manifest.json` file and should be managed in the Admin UI. An extension should only be activatable from the Admin UI for safety reasons. Once active, extensions can execute JavaScript in the Admin UI.

There should be a library package by PocketBase to do common things, like adding new buttons on the `main-menu` to the left.

Server-side extensions should be as is with Go/ JS extensions.

Isos9:
Hi,

https://interval.com/

may be good for inspiration for this feature

timseriakov:
https://docs.medusajs.com/learn/basics/admin-customizations 
https://docs.medusajs.com/resources/admin-widget-injection-zones
also may be good for inspiration for this feature

fayazara:
Yes, please!

A better way to look at this would be an integrations/plugins platform

1. Stripe
2. Zapier
3. Open AI
4. Claude

And these get exposed in the JS SDK to use in the client

CC: @ganigeorgiev

eazuka:
[...]

PR #1561, opened 2023-01-11
Add Telegram auth

iamelevich:
This pull request will add Telegram Auth for [Web App](https://core.telegram.org/bots/webapps) or with [Login widget](https://core.telegram.org/widgets/login).

This is not OAuth2, but I need it for my telegram webapp, so I decided to integrate it inside pocketbase instead of adding new route, cause it can be useful for others I hope.

Please take a look. Still have some things to do:
- [ ] Update js-sdk to support telegram auth
- [ ] Update dart-sdk to support telegram auth
- [ ] Update documentation


ganigeorgiev:
Sorry, I don't want to merge it.

If Telegram doesn't have OAuth2 api then this cannot be part of the core package. It will be just hard to maintain since I don't use Telegram and I can't provide support for it.

You can structure it as server-side "plugin" and create a standalone Go package that accept `core.App` and use that to bind to the `OnBeforeServe` hook and others to register your custom endpoint(s).

_Eventually in the future, we may also add support for [Admin UI addons](https://github.com/pocketbase/pocketbase/issues/898) but for now this a low priority._

iamelevich:
Created a plugin if someone need such functionality. https://github.com/iamelevich/pocketbase-plugin-telegram-auth

PR #1643, opened 2023-01-19
Feature: Gitea OAuth

hlanderdev:
Hi! This PR adds [Gitea](https://gitea.io/en-us/) as an OAuth provider (Confusing, I know, since a provider called _Gitee_ was just added).

A few notes about this PR: Support for scopes in Gitea was just added in this [commit](https://github.com/go-gitea/gitea/commit/de484e86bc495a67d2f122ed438178d587a92526), but it seems it has not hit release as of yet, hopefully soon. Also, I was unsure of what to put as the default Url's, as Gitea is largely meant for self-hosting. For now I put https://gitea.com/ but this seems to be meant for development of GItea itself.


ganigeorgiev:
Thank you! I've squash merged it in the `develop` branch (https://github.com/pocketbase/pocketbase/commit/4c73e16f5404383ed73aa13d4643b477f68ff48e) and it will be available with the next v0.12.0 release (next week or the week following it).

PR #1665, opened 2023-01-22
Prevent autoclosing multiselect choicebox

hibatos:
Allow to select multiple elements in a multi select box, without the select box closing upon each selection.

The proposed change will only add the _closable_ class to dropdown items, if the select is a single select (multiple === false). This prevents the _Toggler_ instance from closing the dropdown, since _handleClickToggle_ checks if the element has the _closable_ class before closing the dropdown.


ganigeorgiev:
Thanks for the suggestion. I may consider applying it in the develop branch during the ongoing relation picker refactoring.

PR #1674, opened 2023-01-24
Use consts for action in broadcastRecord

jungaretti:
This PR adds a const for each action in broadcastRecord. For more information, see https://github.com/pocketbase/js-sdk/issues/136.


ganigeorgiev:
Thanks, I'll consider defining the actions as constants in the `develop` branch (other changes to the realtime apis are also planned for v0.13/v0.14).
Technically the action could be anything, but I guess it doesn't hurt to have the common ones as const close to each other.

Issue #1727, opened 2023-01-31
mailer.Message address fields should be multi-value


cweagans:
Per RFC 5322, To, BCC, and CC fields should be a list of addresses (https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.3).

Currently, they are defined like so:

```go
type Message struct {
	[...]
	To          mail.Address
	Bcc         []string
	Cc          []string
	[...]
}
```

All three of those fields should be `[]mail.Address` instead, I think.

PR #1729, opened 2023-01-31
Add Lua-based cloud code functionality

cweagans:
I've been working on building a cloud code system in the style of Parse Server for Pocketbase.

Here's what I have so far:

* A way to load some Lua scripts given the path to the main script (`--cloudCode /path/to/main.lua`)
* A way to run some Lua just prior to the main script to help with sandboxing or pre-loading modules or whatever (in the event that somebody is e.g. running untrusted cloud code or wants to limit the functionality exposed to developers)
* A Pocketbase API that is exposed to cloud code through the `pb` global, including a handful of modules which are accessible as children of the `pb` global:
    * `pb.meta`: Not sure this should continue to exist, but it was a simple starting point. The only thing available here is `pb.meta.version`, which is just `pocketbase.Version`.
    * `pb.events`: I imagined this as a way to register events without the `app.On*` function calls in Go. Right now, this module reflects all of the `On*` methods out of `core.BaseApp` and exposes the equivalent Lua functions (examples below)
    * `pb.mail`: Since it is not currently possible to send mail messages without writing Go, this is pretty useful. Right now, `pb.mail.send()` just sends a hardcoded message. I'll build out this module further after #1727 is wrapped.
    * `pb.db`: This is just a placeholder for now. My intent is to mirror as much of `app.Dao` as is feasible.

Some of the things still on my list:

* `pb.events`: some events really need the event object passed to the handler function. Right now, I'm not passing it along at all, as I was more interested in getting the reflection stuff working properly + being able to bind Lua functions to Pocketbase hooks.
* `pb.events`: it might make more sense for this to be called `pb.hooks` instead.
* Build some way of auto-generating a `pb.def.lua` file sort of like the files in https://github.com/CppCXY/emmylua-openresty-api so that Emmylua (the Lua language server) will be able to provide cloud code autocompletion and such.
* Provide some way of testing cloud code
* Write Go tests as needed
* Write an example sandboxing script for users running untrusted cloud code. https://stackoverflow.com/questions/1224708/how-can-i-create-a-secure-lua-sandbox and http://lua-users.org/wiki/SandBoxes are both things that I've been looking at.
* Documentation 
* Some kind of work queue functionality. I noticed there was a TODO elsewhere in the pocketbase code to add something like this. I think it would be extremely helpful to have something like this to rely on for e.g. sending lots of email messages. It'd be a lot quicker to queue them up and process them in the background, rather than send them synchronously.
* Build a roadmap for any other modules that should be implemented. There are many gopher-lua extensions available that might be of interest: https://github.com/yuin/gopher-lua#libraries-for-gopherlua
* Registering new routes. I had two different registration methods in mind -- one for static routes, and one for dynamic routes. Static routes essentially just point at a directory of files that are served directly (without processing Lua code on each request). Dynamic routes would invoke the Lua callback for each request.
* Cloud functions/jobs. I thought it was kind of a strange concept when I first started using Parse Server, but I definitely had a couple of uses for them. Relevant Parse Server docs for [functions](https://docs.parseplatform.org/cloudcode/guide/#cloud-functions) and [jobs](https://docs.parseplatform.org/cloudcode/guide/#cloud-jobs).
* Maybe some way to read/write config values? This could be useful for version controlled configuration deployed as part of a CI workflow.
* Migrations and/or [defined schema](https://docs.parseplatform.org/defined-schema/guide/) functionality. I saw that the `jsvm` plugin had migration functionality. At the very least, it might be helpful to re-implement that in Lua. I'd also be very interested in a defined schema mechanism that essentially does the same stuff as the Export/Import schema functionality in the admin UI, but by using a JSON file on disk instead.
* Some kind of audit-trail functionality in the admin UI. Many CMSes have similar functionality ("so and so edited record asdjfsjjdsfkdsj" or "adminuser configured google authentication" or whatever). It would be possible to wire up a `pb.log` module that can write to this log as well.

Here are a couple of super simple cloud code scripts that I've been using during testing:

(you can drop each of these into some random lua file and run them on pocketbase startup with the `-cloudCode` flag)

```lua
--- you can use any of the app.On* functions here -- it's just a lowercase o instead.
pb.events.onAfterBootstrap(function()
    --- this shows up in the pocketbase log output
    print("on after bootstrap")

    --- I tried this after configuring mailhog locally and it ran pretty quick.
    local i = 0
    repeat
        print("sending message:", i)
        pb.mail.send()
        i = i + 1
    until( i == 100 )
end)
```

```lua
pb.events.onBeforeBootstrap(function()
    print(pb.meta.version)
end)
```

I'll update this PR as I continue (or you can close it if you'd prefer that I open a different PR down the road a bit).

Also: I did see the `jsvm` plugin, but I wasn't too excited about extending that: I just really dislike writing Javascript in general. I find Lua to be a very pleasant language + it's a low barrier to entry for people who are using Pocketbase for non-web things (e.g. Flutter applications).

My first attempt at this was using Tengo, but the language tooling really isn't there (e.g. there's no syntax highlighting in Jetbrains IDEs). If you're curious, that branch is here: https://github.com/cweagans/pocketbase/tree/cloudcode -- I'm not planning on continuing with that though.

### DIFF
--- a/examples/base/main.go
+++ b/examples/base/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"github.com/pocketbase/pocketbase/plugins/cloudcode"
 	"log"
 	"os"
 	"path/filepath"
@@ -34,6 +35,22 @@ func main() {
 		"automigrate",
 		true,
 		"enable/disable auto migrations",
+	)
+
+	var cloudCode string
+	app.RootCmd.PersistentFlags().StringVar(
+		&cloudCode,
+		"cloudCode",
+		defaultCloudCode(),
+		"the path to user-defined cloud code",
+	)
+
+	var cloudCodeInit string
+	app.RootCmd.PersistentFlags().StringVar(
+		&cloudCode,
+		"cloudCodeInit",
+		"",
+		"the path to a lua file run before the main cloud code. useful for sandboxing or pre-loading libraries.",
 	)
 
 	var publicDir string
@@ -70,6 +87,9 @@ func main() {
 		Dir:          migrationsDir,
 	})
 
+	// register cloud code.
+	cloudcode.MustRegister(app, cloudCode, cloudCodeInit)
+
 	app.OnBeforeServe().Add(func(e *core.ServeEvent) error {
 		// serves static files from the provided public dir (if exists)
 		e.Router.GET("/*", apis.StaticDirectoryHandler(os.DirFS(publicDir), indexFallback))
@@ -88,4 +108,12 @@ func defaultPublicDir() string {
 		return "./pb_public"
 	}
 	return filepath.Join(os.Args[0], "../pb_public")
+}
+
+func defaultCloudCode() string {
+	if strings.HasPrefix(os.Args[0], os.TempDir()) {
+		// most likely ran with go run
+		return "./pb_cloudcode/main.lua"
+	}
+	return filepath.Join(os.Args[0], "../pb_cloudcode/main.lua")
 }
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/pocketbase/dbx v1.9.0
 	github.com/spf13/cast v1.5.0
 	github.com/spf13/cobra v1.6.1
+	github.com/yuin/gopher-lua v1.1.0
 	gocloud.dev v0.28.0
 	golang.org/x/crypto v0.5.0
 	golang.org/x/net v0.5.0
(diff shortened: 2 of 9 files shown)


ganigeorgiev:
Thank you for working on this but for now I don't plan merging it in the main repo.

I don't think many developers are familiar with the Lua syntax and in general it will be hard to maintain, document and test while also keeping it in sync with the Go and JS code.

The goja integration (aka. the `jsvm` plugin) was added primarily to support external migrations and eventually in the future we may also allow binding to the app event hooks, but goja handles the heavy lifting for us due to its almost seamless interop and the expected maintenance burden should be fairly small.

cweagans:
No worries! Would you like to see it again when it's further along? I plan on developing and using this for my own projects either way, but if you want it in the main project when it's done, I'm happy to share.

ganigeorgiev:
There is nothing wrong exporting it in its own repo as a standalone package that will accept a `core.App` instance so that users can embed it in their own binaries if they want to use it. Later, depending on the demand and how users are using it, we can always make it part of the prebuilt executable (aka. [example/base](https://github.com/pocketbase/pocketbase/blob/master/examples/base/main.go)), but for now I want to avoid the maintenance burden that may come with it.

amleonc:
Now that the JavaScript path didn't end well, I think it'd be a good idea to reconsider this proposal. Lua is more ubiquitous, and as a language is way simpler. I know there are other features higher on the list of priorities, so it's totally understandable if this particular feature isn't worked on in the short - medium term, nonetheless this PR has some valuable ideas. As the maintainer stated in #2825 :

> Lua itself is not that well known as a language, especially by mobile and web developers.

Which is true, but one question is worth asking: Is it better to add extensibility with a somewhat complex popular language (JS), or with a minimal and simple language (Lua)?

Best regards.

Issue #1675, opened 2023-01-24
Feature Request : Okta Oauth2 support / Generic oauth2 support


AlexGiroud:
Hello,

It would be nice if pocketbase would support Okta oauth2 as an authentication provider or a more generic way to add an Oauth provider in the admin pannel.

https://developer.okta.com/docs/reference/api/oidc/

ganigeorgiev:
I don't plan working on this, but users contribution for new OAuth2 providers are welcomed. If some one want to contribute, you can use the latest merged Gitea and Authentik integrations:

- Gitea #1643: https://github.com/pocketbase/pocketbase/commit/4c73e16f5404383ed73aa13d4643b477f68ff48e
- Authentik #1377: https://github.com/pocketbase/pocketbase/commit/6d08a5f36fb92ac00409c9405a7c8d0d5f9f18d5

_(the generated `ui/dist` files can be ignored)_

Support for creating external OAuth2 providers from the Admin UI is not planned because each integration requires some manual mapping (and sometimes even more than 1 request to fetch the necessary user data) and this will differ from one provider to another.

PR #1806, opened 2023-02-07
Okta Oauth2 Provider support

AlexGiroud:
Adding Okta Oauth2 to the providers supported.
#1675


ganigeorgiev:
Thank you for working on this.

I'll review it in more details sometime later this week, but at first glance here some suggestions:

- Remove the debug print statements

- Since Okta is also a hosted service, we can set the default provider urls pointing to it and allow users to specify custom ones in the UI if they are self hosting the platform (aka. similar to the gitlab provider)

- Okta [userinfo response](https://developer.okta.com/docs/reference/api/oidc/#response-example-success-8) seems to return also a `email_verified` field that we have to map and **assign the `AuthUser.Email` only if it is `true`**.
Additionally, we can also make use of the `nickname` field and assign it to `AuthUser.Username` (_if that is what `nickname` is supposed to be; if not we can skip it_).

AlexGiroud:
Hello @ganigeorgiev,

Thanks for the feedbacks (sorry for the debug prints)

- If i remember correctly, there is no default urls for Okta, every org has a custom one, even the test instances.
- For the userinfo, i will add the check for the email. 
- The nickname is not always provided, i can add a check (if provided -> Username, else name -> Username) 

I will try do submit the changes tomorrow

ganigeorgiev:
After the recent comment in https://github.com/pocketbase/pocketbase/pull/1377#issuecomment-1418905209 and the recent Nextcloud app OAuth2 PR - https://github.com/pocketbase/pocketbase/pull/1821, I think it may be better to implement a generic OIDC provider (or as some users suggested to just rename the existing Authentik provider to OIDC) because the OIDC spec has list with standard [scope claims](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims) and [UserInfo fields](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) that should be the same across all OIDC providers.

I'll think a little more on it after the ongoing "View" collection type.

AlexGiroud:
I've just tested with the authentik provider with our Okta, it's working just fine.
It a good plan 
(Could even implement the .well-known/oauth-authorization-server  to fetch the differents url reducing the number of fields to fill in the ui and an advanced toggle to fill the urls manually)

Issue #55, opened 2022-07-10
Simplify the OAuth2 authentication flow


ganigeorgiev:
PocketBase uses the OAuth2 PKCE flow but due the stateless nature of the application and the requirement to support multiple platforms (web, android, ios, etc.) in a similar manner, the current implementation is a little verbose and some users have reported to find it cumbersome to implement in their app (see existing [Auth via OAuth2 guide](https://pocketbase.io/docs/manage-users/#auth-via-oauth2)).

Further more, the current approach has one more drawback - it will not work out of the box with platforms that may not support redirect urls or deep links (eg. AdobeXD and Figma plugins; check similar issue in [Presentator #178](https://github.com/presentator/presentator/issues/178#issuecomment-890531578))

The best solution that I can think of at the moment would be to start a **persistent connection** (eg. SSE) with the client and handle everything in a single call without even reload to be necessary, eg. for web platforms using the JS SDK it could look like this:
```js
client.Users.authViaOAuth2(
  "google",
  function (googleLoginUrl) {
    // open in a popup window or a new tab
    window.open(googleLoginUrl, "_blank");
  },
  function (authData) {
    // success callback...
  }
)
```

Feedback and suggestions for other approaches are welcomed.

_This currently is a low priority, but we'll have to decide on the implementation before v1.0.0 to avoid introducing breaking changes._

burggraf:
Low priority, agreed, but if this works, it'll simplify the OAuth workflow and provider a better DX.  Cool.

Glukoosi:
This would be a good addition since it seems like it is not possible to get redirect url to work with `svelte-spa-router` or other `spa` routers (as you stated at the issue).

jjmountain:
Could we also have a way to add access scopes as URL parameters as part of the redirect URL? 

Currently I'm having to do this in my code:

```
if (provider.name === "google") {
      const updatedUrl = new URL(authUrl);

      const googleOAuthScopes = [
        "https://www.googleapis.com/auth/userinfo.email",
        "https://www.googleapis.com/auth/userinfo.profile",
        "https://www.googleapis.com/auth/forms.currentonly"
      ];
      const revisedScopes = googleOAuthScopes.join(" ");

      updatedUrl.searchParams.delete("scope");
      updatedUrl.searchParams.append("scope", revisedScopes);
      updatedUrl.searchParams.delete("redirect_uri");
      updatedUrl.searchParams.append("redirect_uri", "");
[...]

PR #1821, opened 2023-02-10
Add nextcloud OIDC provider

nat212:
Nextcloud has an app that allows you to use your instance as an OpenID Connect provider (https://github.com/H2CK/oidc/wiki/User-Documentation). I personally use this to manage access to my personal apps, and having this in pocketbase would be useful!


ganigeorgiev:
Thank you for working on this but I'm not sure if it will be a good idea to integrate it in the main package because this doesn't seem to be an officially Nextcloud supported OAuth2 app, right? Also I couldn't find anywhere whether they support the PKCE flow or not (this is not required and #55 will help further, but it would be nice if it is documented somewhere).

Additionally, we have to check whether the provider will always return the email even if it is not verified because in this case we cannot trust it and it would be better to not set since it is used for accounts linking. Usually some providers have `email_verified/emailVerified` or similar field that identifies the `email` state but I'm not sure if that's also the case with this provider.

ganigeorgiev:
Additionally, Nextcloud seems to have their own OAuth2 integration without the need of the external app - https://docs.nextcloud.com/server/16/admin_manual/configuration_server/oauth2.html, but the documentation is very limited and I'm not sure if the above will work with it (there is no information on the scopes or the responses).

PR #1836, opened 2023-02-12
Toggle columns keeps filter if you change collection

gechternachtv:
This is just a one line change to fix a small bug, i added the loadStoredHiddenColumns function call inside the change collection event, because the filter list was persistent when changing collections, but the filter wasn't working until you toggle and un-toggle some element.


ganigeorgiev:
> filter list was persistent when changing collections, but the filter wasn't working until you toggle and some element.

I'm not able to reproduce this. Could you elaborate a little more or provide an example?

ganigeorgiev:
I was able to reproduce it.

I understand what is the ussue, but there is another underlying bug that causes this one.
It should fix it if we reorder the `$: if (collection?.id)` reactivity check before the `$: visibleFields` assignment. I'll submit a fix sometime later today.

ganigeorgiev:
@gechternachtv Should be fixed in [v0.12.3 release](https://github.com/pocketbase/pocketbase/releases/tag/v0.12.3).

Feel free to let me know if you are still experiencing the issue after the upgrade and I'll investigate it more thoroughly.

PR #1847, opened 2023-02-13
Validate file MIME type before size

arslanastral:
Small enhancement to check file MIME type before it's size. Currently, if you upload a file that is not allowed and is over the size limit there is a size limit error first. This can potentially result in a user resizing their file only to later found that the MIME type is also not allowed. We can display more than one error message but for now, I've just swapped the order of the checks.


ganigeorgiev:
Thank you, but I don't think this is necessary.
The mimetype check is more expensive and it is better to be after the size validation.

> This can potentially result in a user resizing their file only to later found that the MIME type is also not allowed. 

You can always "enforce" a mimetypes filter during file upload on client-side level for better UX (_for web there is the `type` file input attribute_).

PR #1853, opened 2023-02-14
added a new line in line 10

ekottifiok:
from this //go:embed all:dist to // go:embed all:dist caused this error when installing or running a file
$GOPATH/pkg/mod/github.com/pocketbase/pocketbase@v0.12.3/ui/embed.go:10:12: pattern all:dist: no matching files found

### DIFF
--- a/ui/embed.go
+++ b/ui/embed.go
@@ -7,7 +7,7 @@ import (
 	"github.com/labstack/echo/v5"
 )
 
-//go:embed all:dist
+// go:embed all:dist
 var distDir embed.FS
 
 // DistDirFS contains the embedded dist directory files (without the "dist" prefix)


ganigeorgiev:
No, it should remain as it is, otherwise it will be interpreted as regular comment and will not embed the resource. See for example https://pkg.go.dev/embed.

I'm not able to reproduce the reported error. My guess is that you are using an older Go version. Note that the `all:dist` embed syntax was added in Go 1.18+.
Please try to upgrade your Go installation and test again.

If you are still not able to make it work after the above, feel free to open an issue/discussion with reproducible steps and I'll have a look at it.

ekottifiok:
I was using an older version of go
Thanks😊

Issue #19, opened 2022-07-08
Responsive Dashboard 


mpldr:
The dashboard is currently not responsive which makes working with it a pain on phones. It would be great to have it adapt to the device.

PR #1867, opened 2023-02-17
added responsive style

fabioars:
Hi folks!

I made some layout adaptations for smaller screen devices. I hope you enjoy it.
Here is some screenshots:


ganigeorgiev:
Thank you, but for now I don't think is necessary. You can also check https://github.com/pocketbase/pocketbase/issues/19.

As mentioned in the above issue, the ui will require major changes in order to make it usable on small mobile devices, not just a responsive version - different table layout, more API rules interface and autocomplete, etc., and for now this remains out of the scope of v1.

PR #1913, opened 2023-02-25
Limit fields return by record list api

luuhai48:
Hi!
I really need feature that you can select the fields returned by record list api. But after having Q&A, they are said this feature is not in the roadmap.
I gave it a try and modify it myself. It is kinda work now, but I know its missing a lot more.
Could this feature be available soon, please?
Thank you.

### DIFF
--- a/tools/search/provider.go
+++ b/tools/search/provider.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/pocketbase/dbx"
+	"github.com/pocketbase/pocketbase/models/schema"
 )
 
 // DefaultPerPage specifies the default returned search result items.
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
Thank you for working on this, but for now this feature request remains out of the scope.

It is in my todo and eventually will be added in the future, but [other tasks are with higher priority right now](https://github.com/orgs/pocketbase/projects/2).

The suggested implementation is not ok because:
- it'll work only with the main collection fields (what about the expanded relations?)
- it seems prone to sql injection
- need tests

Also, please note that in the upcoming v0.13.0 release there will be new "view" collection type that'll allow you to create a readonly collection from a subset of another collection(s) fields. It is not the same, but to some extend could be used as a replacement in case you really need this feature.

PR #1943, opened 2023-02-28
Fix user-facing typo

Semnodime:


ganigeorgiev:
Thank you. I've applied the typo fix locally since the `ui/dist` files need to be also generated for the tests to pass. I'll push it sometime later in the `develop` branch with some other UI changes I'm currently working on.

PR #1950, opened 2023-03-01
Improved DateTime UI

babakfp:
Hey

Note: I don't know how these changes will play with other features/parts/elements of the Flatpickr component. For example, what if a seconds picker gets added to the bottom bar of the date picker?

After:

after

Before:

before


ganigeorgiev:
Thank you, but I'm not sure about the change since it is not clear that the month and year fields are input/select form elements.

I'll consider updating the existing UI input styles to look a little more refined and polished, but for now is a low priority.

PR #1954, opened 2023-03-01
Apple OAuth2 (need help with testing!)

### LINKED ISSUE #202: [Need help with testing] Feature request: Apple Signin

ganigeorgiev:
This is a refactored version of #899 PR.

**Since I don't have Apple Developer account to verify that it works, I'll appreciate if someone has the time to give it a try locally and test the integration.**

To do that:

0. [Install Go 1.18+](https://go.dev/doc/install)

1. Clone the [`apple-oauth2` branch](https://github.com/pocketbase/pocketbase/tree/apple-oauth2)
  ```sh
  git clone --single-branch --branch apple-oauth2 https://github.com/pocketbase/pocketbase.git
  ```

2. Navigate to `examples/base`

3. Run `CGO_ENABLED=0 go run main.go serve`

4. Copy/paste the 2 static `pb_public/index.html` and `pb_public/redirect.html` files as shown in https://pocketbase.io/docs/authentication/#web-oauth2-integration.

5. Enable the Apple OAuth2 provider from the "Admin UI > Settings > Auth providers"
   You can follow this tutorial from Okta how to generate the various required provider keys - https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple#how-sign-in-with-apple-works-hint-it-uses-oauth-and-oidc
   Note that the Admin UI comes with secret generator, so you don't have to generate it manually via ruby as shown in the above guide.
   You need:
   - `client_id` (aka. "Apple's Services Id")
   - `client_secret` which could be generated from the Admin UI when provided:
     - `client_id`
     - `team_id`
     - `key_id`
     - `private_key` (not stored on the server; used only for the secret generation)

6. Configure the Apple OAuth2 app. For `redirect_uri` use `https://yourdomain.com/redirect.html`. Note that per the Apple docs it must be a valid domain with `HTTPS` schema. You can use a ssh tunneling service like [ngrok](https://ngrok.com/) to create a public address for your localhost.

   Additionally, if you are manually constructing the sign-in url and not using the one from `listAuthMethods()` call, you have to set the query param `response_mode=query`.

7. Navigate to [`https://127.0.0.1:8090`](https://127.0.0.1:8090) and click on the "Login with Apple" link.

8. Report any errors.


jonafeucht:
LGTM! 🙌

Works with the provided web OAuth2 implementation.
I’ll have to spin up a nodejs server and test it with a Flutter app, but from what I’m seeing it’ll most likely work.
I’ll only find time over the weekend for this though, ofc I’d welcome it if someone finds the time before that.

tools/auth/apple.go
```diff
@@ -0,0 +1,221 @@
...
+
+func (p *Apple) parseAndVerifyIdToken(idToken string) (jwt.MapClaims, error) {
+	if idToken == "" {
+		return nil, errors.New("empty id_token")
+	}
+
+	// extract the token header params and claims
+	// ---
+	claims := jwt.MapClaims{}
+	t, _, err := jwt.NewParser().ParseUnverified(idToken, claims)
```
jonafeucht, line 108: Should we explicitly check if the token is signed with the expected algorithm or is this already done on Apple's end?

jonafeucht (commented):
see comments

tools/auth/apple.go
ganigeorgiev: This is done during the actual verification below (#l154)

forms/apple_client_secret_create.go
```diff
@@ -0,0 +1,112 @@
...
+	)
+}
+
+// Submit validates the form and returns a new Apple Client Secret JWT.
+func (form *AppleClientSecretCreate) Submit() (string, error) {
+	if err := form.Validate(); err != nil {
+		return "", err
+	}
+
+	signKey, err := parsePKCS8PrivateKeyFromPEM([]byte(strings.TrimSpace(form.PrivateKey)))
```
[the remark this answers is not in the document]
ganigeorgiev: Thanks. Yes it should be the same. I somehow missed it yesterday when was searching through the jwt docs.

PR #1966, opened 2023-03-04
Resize table columns

mjadobson:
As per https://github.com/pocketbase/pocketbase/discussions/1542

chrome-capture-2023-2-4

It uses a fairly generic svelte action, so it doesn't interfere much with the existing code. Took a bit of time to smooth out any bugs, but seems stable now.

Thanks again for your work on the project; the latest release was great 👍.


ganigeorgiev:
I'm not sure about this feature.

I haven't tested it locally, nor reviewed the code changes, but just from the screenshot it looks kindof strange since the table is no longer 100% width (maybe because of the fixed layout?).

mjadobson:
> I'm not sure about this feature.
> 
> I haven't tested it locally, nor reviewed the code changes, but just from the screenshot it looks kindof strange since the table is no longer 100% width (maybe because of the fixed layout?).

I did think this after submitting the pull request with the gif.

It makes the implementation slightly less elegant, but I can keep a minimum width for the table. I will revise and see what you think.

ganigeorgiev:
I don't think `table-layout: fixed` is necessary and it could cause some issues with adding for example a new column after the resizing.

Additionally, there should be something that will prevent resizing below some min-width threshold, because from the screenshot the id seems to be cropped and I'm not sure if this is a good idea.

mjadobson:
I've updated the code to keep the minimum table width:

ezgif-1-ceb5fe61a4

This implementation does require a fixed table layout. Trying to keep columns a specific width and the resize-handle tracking the cursor was problematic with a fluid layout.

The columns will clip as you resize. I personally prefer this to min widths.

Let me know what you think, feel free to close if it doesn't make sense for the project.

ganigeorgiev:
Sorry, I appreciate the work you've put into this but I don't want to rush it and merge something just because there is a PR for it.

I like the idea of resizable columns but the current implementation feels a little brittle and I'm not confident that we are handling all edge cases (eg. on windows small->big resize I guess we need also to bind to the resize event to recalculate the initial table width? I'm also not sure how it will behave if we do external layout changes, eg. via plain css in a future responsive version or just toggling sibling dom elements, etc.).

I've added it in my local todo to search for other options and eventually something similar could be implemented in the future, but I don't want to invest time into it right now and for now remains out of the scope.

PR #1969, opened 2023-03-04
Feature/add hubspot auth

HT808s:
The svg for Hubspot is missing from the remixIcon library

I made a request for it but it's still pending
https://github.com/Remix-Design/RemixIcon/issues/494


tools/auth/hubspot.go
```diff
@@ -0,0 +1,92 @@
...
+				Name    string `json:"name"`
+				Primary bool   `json:"primary"`
+			} `json:"teams"`
+		} `json:"results"`
+		Paging struct {
+			Next struct {
+				After string `json:"after"`
+				Link  string `json:"link"`
+			} `json:"next"`
+		} `json:"paging"`
```
ganigeorgiev, line 67: We don't need to map every field from the response. Map only those fields that are directly used in the `*AuthUser` struct (`id`, `userId`, `email`, `firstName`, `lastName`).

_Side-note: What is the difference between `id` and `userId`?_

ganigeorgiev (changes requested):

HT808s:
@ganigeorgiev I would like to add options to customize the scopes of the Oauth URL, can I do it?
If yes, how should I do it?

ganigeorgiev:
>  I would like to add options to customize the scopes of the Oauth URL, can I do it?

I don't think it is necessary (at least at this stage) because users can change the default scopes in the sign-in/sign-up url string returned by `listAuthMethods()`.

tools/auth/hubspot.go
HT808s: > Side-note: What is the difference between id and userId?

In Hubspot, users and owners are separate resources, but an owner is also a user. (Weird design)
Here `id` is the owner id since it is sourced from the owner endpoint

From what I know, assigning users to a deal, a contact or a company, the owner id is used
https://community.hubspot.com/t5/APIs-Integrations/User-id-vs-owner-id/m-p/231212

ganigeorgiev: Hm, I don't have experience with HubSpot but maybe the `owners` endpoint is not the correct one here. From the docs: 

> HubSpot uses owners to assign CRM objects to specific people in your organization. The endpoints described here are used to get a list of the owners that are available for an account.

Which as far as I understand, more than 1 users/owners could be returned. How we decide which to assign? At the moment we always will extract the first one.

Additionally, maybe we should stick to only the `"oauth"` scope from https://developers.hubspot.com/docs/api/working-with-oauth#scopes.

tools/auth/hubspot.go
```diff
@@ -0,0 +1,78 @@
...
+// Hubspot allows authentication via Hubspot OAuth2.
+type Hubspot struct {
+	*baseProvider
+}
+
+// NewHubspotProvider creates a new Hubspot provider instance with some defaults.
+func NewHubspotProvider() *Hubspot {
+	return &Hubspot{&baseProvider{
+		scopes: []string{
+			"crm.objects.owners.read",
```
ganigeorgiev, line 25: Maybe change to `"oauth"`?
_(https://developers.hubspot.com/docs/api/working-with-oauth#scopes)_

ganigeorgiev:
@HT808s I think the implementation needs a little more clarification regarding the HubSpot apis and its terminology.

In my opinion [`GET /account-info/v3/details`](https://developers.hubspot.com/docs/api/settings/account-activity-api) should be the endpoint for the user info (aka. `userApiUrl`) but it doesn't seem to return the `userId` (it returns `portalId`).

HubSpot is not a generic OAuth2 provider and its use case seems too business specific.

Could you elaborate what is your use case and how do you plan to use the integration?

tools/auth/hubspot.go
HT808s: You make good points, so I searched the internet for an answer
I think I've found something [here](https://community.hubspot.com/t5/APIs-Integrations/Get-current-user-id-and-info-from-oAuth-connection/m-p/240602)

Now the only problem is that the URL has a parameter 😅
https://legacydocs.hubspot.com/docs/methods/oauth2/get-access-token-information

HT808s:
> 

Indeed it's not a generic oauth provider tailored to a single user

> Keep in mind though that apps are authed for an entire portal, so all users in a given portal can use the features of your integration. There isn't currently a public user API, so it's not possible to get details for all users in a portal

My goal is to create a public app for Hubspot using Pocketbase as my backend
I thought it would be cleaner to integrate it as a native provider, but actually it's may not be the same use case as PB

tools/auth/hubspot.go
ganigeorgiev: For the parameter in `GET /oauth/v1/access-tokens/{token}` we can specify a custom `FetchRawUserData` function for the HubSpot provider, something like:

```go
func (p *Hubspot) FetchRawUserData(token *oauth2.Token) ([]byte, error) {
    req, err := http.NewRequest("GET", p.userApiUrl + token.AccessToken, nil)
    if err != nil {
        return nil, err
    }

    return p.sendRawUserDataRequest(req, token)
}
```

ganigeorgiev:
I'm not really sure what would be the best approach here.

I have no experience with HubSpot and I'm not familiar how they manage their users.

Even if we end up implementing `GET /oauth/v1/access-tokens/{token}` as user info endpoint, still I'm not sure how this user could be actually used in combination with PocketBase. What if this user is not part of a "portal" or the "owners" group, but some random HubSpot account?  What you can do with such user and what would be the point of having an equivalent PocketBase auth record?

To some extend, the use case seems similar to the existing `Livechat` provider, but at least there the accounts/users don't have so much levels and grouping and it is a lot easier to distinguish and generalize, while in HubSpot the API seems to be very tightly integrated to their app ecosystem and the OAuth app is more of a standard to authorize for their APIs, rather than being an auth provider.

ganigeorgiev:
@HT808s I'm closing this PR for now since I don't think it will be a good idea to have it integrated in the main package.

I don't feel confident enough that I understand how HubSpot manage their users. Outside of that HubSpot doesn't seem to support the PKCE flow, the use case seems to be very niche and it is maybe better the integration to be left outside of PocketBase because I guess you still have to somehow distinguish the different type of users, contacts, etc. and the generic API with just the `AuthUser` response that all other providers currently implement may not be sufficient.

Feel free to comment or reopen a new PR if you have a more clear idea and understanding of the HubSpot APIs, but for now this remains out of the scope.

PR #2195, opened 2023-04-01
Decode url in the logs page for better view

Uziniii:


ganigeorgiev:
Thank you, but let's keep it for now as it is.

I'm not sure if it will be a good idea because users will not be able to differentiate the missing/invalid encoded url query parameters if we only display the decoded version.

Alternatively we can show both the normalized and the raw urls but the logs UI will change anyway with the planned refactoring in the near future as part of the [Generalize logs task](https://github.com/orgs/pocketbase/projects/2/views/1?filterQuery=genera&pane=issue&itemId=18223491).

Issue #2238, opened 2023-04-06
Feature Request: Truncate collection


jimafisk:
**Summary:** Bulk delete thousands of records using the Admin UI.

**Details:** We have a couple of collections in our PocketBase instance that have thousands of records. We'd like to be able to completely clear out these records when testing a custom data migration. Currently in the admin dashboard it shows 30 records by default and you can scroll down and load 30 more at a time. It would be nice if the bulk select at the top had the option to select the currently loaded records or all the records in the entire collection, which you then could bulk delete. 

<details>
<summary>Concept for how this could look in the UI</summary>

![all_records](https://user-images.githubusercontent.com/5913244/230457119-6b7a2c45-f612-474a-afa7-0cb7fd1e1a55.png)

</details>

ganigeorgiev:
Something like this was requested previously in https://github.com/pocketbase/pocketbase/discussions/560.

I think it will be confusing to have a "Select all ..." in the bulk bar UI. Instead it might be better to have a "Truncate" option in the collection settings dropdown (eg. next to the "Duplicate" option). The "Truncate" option could trigger a dedicated endpoint accessible only by admins.

I'll add it to the roadmap but for now remains a low priority.

buldezir:
maybe someone will find it helpful:
tmp "hack" to implement it in your code "for" js sdk:
```ts
import { RecordService } from "pocketbase";

declare module "pocketbase" {
    interface RecordService {
        truncate(): Promise<void>;
    }
}

RecordService.prototype.truncate = async function () {
    const old = await this.getFullList();
    for (const oldRecord of old) {
        await this.delete(oldRecord.id);
    }
    console.log(`Truncated ${this.collectionIdOrName}`);
};
```

jonshipman:
My `truncate collection` command

```
$app.rootCmd.addCommand(
  new Command({
    use: 'truncate',
    run(cmd, args) {
      if (!args[0]) return console.error('Need collection name');

      $app
        .dao()
        .db()
        .newQuery('DELETE FROM ' + args[0])
        .execute();
    }
  })
);
```

Does it in on sqlite query instead of how many records you have.

ganigeorgiev:
Option to truncate collection via the web api (superusers only) and programmatically via `app.TruncateCollection(collection)` method was implemented as part of the refactoring and will be available with the upcoming v0.23.0 release soon (_there are no ETAs yet_).

PR #2306, opened 2023-04-16
2238 truncate collection draft

buldezir:
https://github.com/pocketbase/pocketbase/issues/2238

proof of concept (working) as @ganigeorgiev proposed

sorry for my Golang code, first time using it.


ganigeorgiev:
This is slightly more complicated than triggering the truncate sql statement since we have to check for record references before executing the truncate.

We need also new app level hooks (before and after truncate), tests for the dao method (+godoc), tests for the api handler and to move the client `truncate()` method into the JS and Dart SDKs (+ tests for it in each SDK).

PR #2466, opened 2023-05-12
Add accepted mime types to file input html element

Nikhil1920:
Now file explorer will only show files with the accepted mime types for selection by default


ganigeorgiev:
Thanks! To avoid the back-and-forth I've merged the suggestion locally.

Issue #2236, opened 2023-04-06
Feature Request: Embedded SQLite prompt


jimafisk:
This would probably be pretty low priority, but often times I'm deploying PocketBase to an environment that only has the binary and I'd like to avoid installing other dependencies. It would be nice if I could still run SQLite commands since it's already embedded in the binary. I'm picturing being able to run something like `./pocketbase sqlite` would open a prompt that I could then use to execute custom sql commands. Thank you!

eugenepentland:
I made a [PR ](https://github.com/pocketbase/pocketbase/pull/2477)for this request. Let me know if that's what you were looking for. First time contributing to open source code.

figuerom16:
https://gitlab.com/figuerom16/moxylib

I implemented a boilerplate inline SQLite interface that does datatables and exports CSVs.  Someone can copy/change the bits into a Svelte version.

It's using a few plain JS libraries so it might not be the cleanest to switch over:
htmx - embedding the result.
surreal - JS helper.
code-input w/ (highlightjs) - The SQL editor portion.

The repo has hints in the README for where the important bits are, but the repo was made to be as plain as could be.

ganigeorgiev:
New "SQL console" section (_Settings > Debug_) was added in the `develop` branch and will be available with the next v0.39.0 release _(most likely sometime next week)_.

Note that this is intended for one-off analytic queries (e.g. https://github.com/pocketbase/pocketbase/discussions/7638), the occasional `VACUUM`/`PRAGMA optimize` or debug purposes and not as the primary interface for interacting with your PocketBase data because it can break your application if not used with proper care!

PR #2477, opened 2023-05-14
Sqlite CLI Query Command

eugenepentland:
Added a new CLI command called "sqlite". It allows you to run SQLite queries on the database from the CLI.

Example: pocketbase.exe sqlite "DELETE FROM users WHERE name = 'foo'"

I saw issue [2236](https://github.com/pocketbase/pocketbase/issues/2236) and figured it would be easy to implement. First time contributing to open-source code. I ran the linter and there are no errors with my code.

### DIFF
--- a/cmd/sqlite.go
+++ b/cmd/sqlite.go
@@ -6,8 +6,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// NewServeCommand creates and returns new command responsible for
-// starting the default PocketBase web server.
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
Thank you but it is a little more complicated as we'll also to output a table with the query result, eg. in case a `SELECT` or `EXPLAIN` is used.

In any case, for now this is out of the [roadmap](https://github.com/orgs/pocketbase/projects/2) and has a very low priority.

_Before contributing, please check https://github.com/pocketbase/pocketbase#contributing._

PR #2533, opened 2023-05-22
Implement vk oauth2

imperatrona:
hi, i've implemented vk oauth2.
it's all working, but i had to remove pkce options cause vk strangely doesn't support it for server side auth.


tools/auth/vk.go
```diff
@@ -0,0 +1,82 @@
...
+
+// NewVkProvider creates new Vk provider instance with some defaults.
+//
+// Docs: https://dev.vk.com/api/oauth-parameters
+func NewVkProvider() *Vk {
+	return &Vk{&baseProvider{
+		scopes:     []string{"email"},
+		authUrl:    vk.Endpoint.AuthURL,
+		tokenUrl:   vk.Endpoint.TokenURL,
+		userApiUrl: "https://api.vk.com/method/users.get?fields=photo_max,screen_name&v=5.131",
```
ganigeorgiev, line 30: Wouldn't `users.get` returns all users associated to the current VK account? How we decide which one to choose (currently it seems that always the first one will be selected)?

I'm having difficulties navigating through the VK russian docs, but wouldn't [`account.getProfileInfo`](https://dev.vk.com/method/account.getProfileInfo) be more suitable for retrieving the user data?

tools/auth/vk.go
```diff
@@ -0,0 +1,82 @@
...
+		Id:           strconv.Itoa(extracted.Response[0].Id),
+		Name:         extracted.Response[0].FirstName + " " + extracted.Response[0].LastName,
+		Username:     extracted.Response[0].Username,
+		AvatarUrl:    extracted.Response[0].AvatarUrl,
+		RawUser:      rawUser,
+		AccessToken:  token.AccessToken,
+		RefreshToken: token.RefreshToken,
+	}
+
+	email := token.Extra("email")
```
ganigeorgiev, line 76: Is the email guaranteed to be verified? Could you point me to the docs?

tools/auth/vk.go
```diff
@@ -0,0 +1,82 @@
...
+			AvatarUrl string `json:"photo_max"`
+		} `json:"response"`
+	}{}
+
+	if err := json.Unmarshal(data, &extracted); err != nil {
+		return nil, err
+	}
+
+	if len(extracted.Response) == 0 {
+		return nil, err
```
ganigeorgiev, line 63: `err` in this case will be `nil`. Instead we can return something like `errors.New("missing response entry")`, but please check the above comment for `account.getProfileInfo` as I'm not sure that `users.get` is the correct user data endpoint.

ganigeorgiev (changes requested):

tools/auth/vk.go
imperatrona: [`account.getProfileInfo`](https://dev.vk.com/method/account.getProfileInfo) works only for token that was recieved through Implicit Flow, and can't be used on server.

> Этот метод можно вызвать с ключом доступа пользователя, полученным в Standalone-приложении через [Implicit Flow](https://dev.vk.com/api/access-token/implicit-flow-user).

`users.get` indeed returns a current user if `user_ids` is not set.

tools/auth/vk.go
imperatrona: Oh sure, i'm sorry thats my fault.

tools/auth/vk.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	extracted := struct {
+		Response []struct {
+			Id        int    `json:"id"`
+			FirstName string `json:"first_name"`
+			LastName  string `json:"last_name"`
+			Username  string `json:"screen_name"`
+			AvatarUrl string `json:"photo_max"`
```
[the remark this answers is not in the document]
imperatrona: I think `deactivated` field can only be returned if requested user is not current one. If you request data about current user and it's deactivated it should return error at `FetchRawUserData` cause it would be unauthorized. I think it's error [18 - Page is deleted or banned](https://dev.vk.com/reference/errors).

ganigeorgiev:
@imperatrona Thank you! I've just created a VK account and tested the integration and it seems to work fine.

To avoid the back and forth, I've squash merged the changes in the `develop` branch (with some nitpick additions) - https://github.com/pocketbase/pocketbase/commit/af71b63f23e5db5eab965c982c80c1f0c36c8a60.

The provider will be part of the next v0.17.0 release (_there are no ETAs yet_).

PR #2679, opened 2023-06-10
Updated README.md

penevl:
1. Updated how the e-mail address looks 
2. The link auto opens the default mail app with a subject of `Security vulnerability report` and a send address set to `support@pocketbase.io`


ganigeorgiev:
Thank you, but the email address was intentional typed this way to prevent spammers and malicious bots scrapping it.

Issue #1659, opened 2023-01-21
in collections api, recored paggination is not working correctly


alinapso:
on PB v0.11.3 

when calling the api api/collections/<collection-name>/records?page=2&perPage=20
I have 5

 recoreds in the collection, I was expecting to get empty result, but for some reason it is returning the same resualt as ?page=1&perPage=20

![1-20](https://user-images.githubusercontent.com/57593645/213888550-6d3b8bfc-b657-4423-a3ee-ca9658c88d1f.png)
![2-20](https://user-images.githubusercontent.com/57593645/213888552-d449e162-bb5f-4a45-a7aa-81cab7f697dd.png)

ganigeorgiev:
This is working correctly. The `page` query parameter is normalized and automatically "capped" to the max possible number so even if you provide 99999 for example it will always return the `totalPages` page results (in your case that is 1).

Issue #1232, opened 2022-12-11
Proposal: Custom rate limiting rules


MarvinJWendt:
Preliminary: I know that rate limiting has been proposed before (in #387), but due to the lack of a description, I would like to submit a justified proposal herewith.

## Description

I propose the option to implement basic rate-limiting to specific API endpoints. Most, if not all apps, would benefit from rate-limiting user and record creation. For this proposal, I want to create an imaginary app that is like Instagram.

### Attack vector

Servers / apps will, sooner or later, be attacked. 

1. A common attack vector is to spam the creation of posts (like pictures on our imaginary Instagram clone). At the current state, PocketBase (in its bare form) would allow the creation of thousands of posts per second. This is obviously nothing that we would want in our app.
2. Another attack vector is dumping the whole database. An attacker could query all posts and increase the "page" in a loop. This would cause intense load on the servers memory and the attacker would have every post locally (which might be millions on big apps).
3. The probably most dangerous attack would be to create thousands of users. Those are hard to delete, as it's not tracked which IP created the user. A simple script is currently capable to create users in a loop. Such an attack could hardly be prevented in PocketBases current form.

### Current solution

The current solution requires the use of a reverse-proxy. This would be feasible, but the programmer would have to create individual rules for each route (which are currently not marked as stable, so the rules might need to be updated frequently). With larger apps this is hardly possible to keep synchronous and would require the usage of lots of wildcards (posts creation might have a different rate-limit than post-view for example). Also, some people do not have a reverse-proxy or are not capable to set one up.

### Proposed solution

I propose that an option is added to each collection, how rate-limiting should be handled. This option could be embedded in the existing API rules. A simple implementation could look like this:

> **Note**
> Text in square brackets could be an input field as dropdown or raw input

```
Allow [user/IP/IP||user] [number] [creations/updates/views/lists] per [number] [second/minute/hour] 
```

> **Note**
> The limiting metric IP||User should count request on the same IP or the same user, to prevent users to use proxy lists.
> When anonymous users have access to the collection endpoint, it should fallback to IP||session_id

As a quick mockup, this could look like:
![image](https://user-images.githubusercontent.com/31022056/206884363-d278cbbd-fed2-4162-b182-500925294e94.png)
![image](https://user-images.githubusercontent.com/31022056/206884774-7579789e-4a8c-4f0f-be59-7a926f5a2cd3.png)

Maybe this could be integrated in the API Rule syntax somehow, but I'll leave that up to you :)

[...]

PR #2749, opened 2023-06-19
Make default pagination item count and maximum item count customizable from admin ui

kubilaysalih:
Hello,

With this pr, the maximum number of records that API can return and the number of default records will be editable via the admin panel.


ganigeorgiev:
This was requested before, but for now is left aside as the default 500 limit should be enough for most cases (_if you want to get all items at once, the SDKs comes with `getFullList()` method that will perform sequential batch requests and will return a final array with all collection records_).

Eventually sometime in the future we may introduce similar configurable limits options per collection level (could be even combined eventually with https://github.com/pocketbase/pocketbase/issues/1232).

_I appreciate the effort, but please don't file PRs for new features without previously discussing the implementation details in an issue or a Q&A discussion, to avoid spending too much time in something that may not get merged. I try to follow a specific plan/[roadmap](https://github.com/orgs/pocketbase/projects/2/views/1) and such PRs come out of nowhere, skew the initial planning and always require some tedious back-and-forth._

kubilaysalih:
I couldn't find anything about pagination settings in roadmap, but I wanted to open pr because of [this comment](https://github.com/pocketbase/pocketbase/issues/1659#issuecomment-1399341030). 

I made such an improvement because pulling 20k or more data with the getFullList is wasted time :/

Thanks for your time.

ganigeorgiev:
> 20k or more data with the getFullList is wasted time 

I'm not sure what is your use case, but please note that you could easily exhaust your server resources when fetching so many items at once and most of the time it would be better to batch or paginate the requests.

kubilaysalih:
You are absolutely right, it is necessary to paginate when pulling 20k data, but I think I should have higher standards for this much data, instead of making requests that I can get 500 records all the time. Right now I can handle this in 4 or 5 requests instead of making 40 requests. I think it's possible to speed up a lot with less requests with sacrificing some resources. All I want to say is that you should leave that desicion to the user. Pocketbase is already sufficient for most scenarios. If I want to fetch 100k record in a request, its my problem, not pocketbase.

PR #2762, opened 2023-06-20
Implement Yandex oauth2

imperatrona:
Hi, I've implemented Yandex oauth2.

I've asked support if they return only verified emails, and their response is that they do.
<details><summary>Support reply</summary>
screenshot of support reply, the text of which provided below

<p>
Which translates as:

Hello!
I'm coming back with information from the developers.
The returned e-mail is confirmed. There is no need to validate it again.

</p>
</details>


ganigeorgiev:
Thank you! I'll review and test it sometime later this week.

ganigeorgiev:
I've tested it locally and squash merged the changes in the [`develop` branch](https://github.com/pocketbase/pocketbase/tree/develop).

The provider will be part of the next v0.17.0 release soon (_I don't have a fixed ETA yet, but if nothing else critical occur, I expect to release it sometime mid-July_).

Issue #159, opened 2022-07-18
Feature Request: Handle "singleton style" content in a better way


kevmodrome:
If you're building something that needs singleton style data you have to create a collection for each such piece of data and then have one single row in there.  For example, if you're using PocketBase to store marketing text for your landing page, etc.

It would be great if this could be abstracted over in some way. It seems like a common enough use-case that it would be nice.

ganigeorgiev:
I agree that this will be nice to have. I'm just not sure what would be the best way to implement it without complicating the API too much.

I think we can keep the current internals (aka. storing each collection and records in their own table), but provide an option in the Admin UI when you create a new collection to choose between a "Single record" or "Multiple records" collection.

Or in other words, I propose to have only different visual representation with minimal server-side changes, so that when the user pick the "Single record" option, in the Admin UI:
- we will create automatically the collection record in the background (this collection will be allowed to have only one record)
- we will directly show the record values on the collection page, instead of the current records table listing
- we will show only the **view** and **update** API rules in the "Single record" collection settings panel, since create and delete will be not allowed (there is always delete option for the entire collection)

Doing this we would preserve most of the existing internals and in the same time provide a slightly better dev experience.
I'll add it to the roadmap but for now it will be after the import/export feature.

kevmodrome:
I think that makes sense. Good ideas. 👍

tryffel:
Wouldn't a simple key-value Resource solve this problem?

ganigeorgiev:
@tryffel I guess it depends on your use case. You can store mostly everything in a JSON field but you would loose the various filtering and validations that comes with the different field types.

I can think of several other use cases for this feature request:
- for simple reporting pages, internally consisting of a single record with several computed/rawsql fields
- for global app settings/options (eg. app colors, various boolean/toggle controls, etc.)

hungcrush:
Oh.. I'm really need this ...

hungcrush:
Hi @ganigeorgiev I'm working on it, 

> we will create automatically the collection record in the background (this collection will be allowed to have only one record)

Could you give me some suggests for this? I will listen after collection is created, if collection type is singleton I will create record automatically



https://user-images.githubusercontent.com/8098916/219852566-840ba515-1716-4be7-8d91-2a744ce0a32b.jpg

jgb-solutions:
+1

upgreidas:
[...]

PR #2778, opened 2023-06-23
Feature/singleton-collections

upgreidas:
Implemented singleton collection type feature. SaveCollection method of Collection DAO now checks if collection is new and if it's type is "singleton" and creates empty record with ID set to "default". It does not trigger any events. I assumed it's not necessary, because prior this, said collection does not exist.

On the UI side created new component named RecordSingleton. Now PageRecords component renders RecordSingleton or RecordsList component based on active collection's type. New component is created based on RecordUpsertPanel component, so there is some code repetition. Might create new, reusable, component like RecordForm or RecordSchemaFields in the future, so we could implement new features, new field types in shared component.


ganigeorgiev:
As mentioned in https://github.com/pocketbase/pocketbase/issues/159, this feature currently is left aside for later consideration.

For now I'll close the issue but in general, if we decide to merge it, we'll need:

- unit and api tests for the backend changes
- some additional checks to prevent record deletion
- some additional checks to prevent creating new records in the collection

For the UI changes I have no strong opinion for now. The Admin UI will most likely will be refactored anyway in relation to #898.

PR #2795, opened 2023-06-26
CLI option to expand autocert host whitelist

pkmollman:
Let's Encrypt cert enrollment has issues when behind NAT.
The service can't bind to the public domain's IP when the host doesn't own it, crashing the process on startup.
When changing --https to bind to a private IP, autocert would trip over the fact that the public domain isn't in the host whitelist.

This should provide a simple method to avoid the bind error, and add any other domains autocert could need to verify.

### DIFF
--- a/apis/serve.go
+++ b/apis/serve.go
@@ -26,6 +26,7 @@ type ServeOptions struct {
 	ShowStartBanner bool
 	HttpAddr        string
 	HttpsAddr       string
+	TlsDomain       string
 	AllowedOrigins  []string // optional list of CORS origins (default to "*")
 	BeforeServeFunc func(server *http.Server) error
 }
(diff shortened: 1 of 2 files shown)


pkmollman:
Thought I made a PR on the wrong branch, and panic closed this. Maybe I didn't?
Glad to get feedback on this.

ganigeorgiev:
_Please consider filing an issue or discussion before investing too much time around a PR that adds new feature(s). `apis.Serve` also has slightly changed in the `develop` branch._

In any case, one possible workaround in your case could be to map the domain in your `/etc/hosts` file.

Also keep in mind that Let's encrypt doesn't allow issuing certificates to bare IPs.

I'll have to think a little more on it as it may adds confusion around the `--http`/`--https` flags.

EthanTuning:
Hello, I am curious if this issue could have something to do with an issue I am currently having. Pocketbase has been running just fine on this Linode server I have set up. Until randomly today I could not connect to Pocketbase API from my application, nor could I access the admin UI either. I stopped the Linux service I had set up and manually ran `pocketbase serve` command to see if I could find anything out. Sure enough when I try to access the admin UI I got this:

Screenshot_1

What I find most odd, is that this started happening after months of Pocketbase running without any issues on this server, and I have not changed anything about the server.

What are your thoughts on how I could mitigate, and would this issue have anything to do with this PR?

EthanTuning:
Hello, I am curious if this issue could have something to do with an issue I am currently having. Pocketbase has been running just fine on this Linode server I have set up. Until randomly today I could not connect to Pocketbase API from my application, nor could I access the admin UI either. I stopped the Linux service I had set up and manually ran `pocketbase serve` command to see if I could find anything out. Sure enough when I try to access the admin UI I got this:

Screenshot_1

What I find most odd, is that this started happening after months of Pocketbase running without any issues on this server, and I have not changed anything about the server.

What are your thoughts on how I could mitigate, and would this issue have anything to do with this PR?

EthanTuning:
@ganigeorgiev I appreciate the reply. I will outline my issue in a QA discussion, and see if we can get it figured out over there. Thank you!

Sagor64:
Nice

ganigeorgiev:
A slightly different version was implemented in the `develop` branch.

Please also check https://github.com/pocketbase/pocketbase/discussions/3190.

PR #2818, opened 2023-06-30
Update SchemaFieldText.svelte

dschissler:
Add escape to help text.


ganigeorgiev:
Thank you, I'll apply the fix locally in the `develop` branch to avoid the rebase and the ui/dist merge conflicts that will result from the change.

PR #2865, opened 2023-07-06
feat: wechat oauth2 provider

aleadag:
Implement WeChat OAuth2. No email address will be returned from WeChat user API. Normally, we will persist the openid/unionid, and interact with user WeChat.

Other things to note:
- WeChat use "appid/secret" instead of "client_id/client_secret" for the parameter name;
- The access token API response with a JSON, but in its head, the content-type is "text/plain". This will make golang.org/x/oauth2 fail to parse the response. I manually fetch the token to workaround this issue.

image


ganeshkbhat:
nice. i love your project @ganigeorgiev . great alternative to than django; while I cannot compare model based api generator and an dynamic rad based api generator. was trying this out and was looking at the interface.  

while, adding oauth for as many social oauths is great. is there an option for generic oauth provider (with appropriate links and callback links) for basic password, token, refresh token. this should ease the usage for non-approved or non-existent oauth providers or their integration since the design for all 4 types of oauth providers will be conceptually the same. if i am being pushy on this, i pitch to contribute to this feature. :)

ganigeorgiev:
@ganeshkbhat I don't see how this is related to this PR. Please if you have different suggestion or question open a new issue or discussion and don't spam existing ones as it adds unnecessary confusion. Additionally, we already have a generic OAuth2 provider and that is the OIDC clients for the cases when the implementation is standardized.

aleadag:
> @ganeshkbhat I don't see how this is related to this PR. Please if you have different suggestion or question open a new issue or discussion and don't spam existing ones as it adds unnecessary confusion. Additionally, we already have a generic OAuth2 provider and that is the OIDC clients for the cases when the implementation is standardized.
> 
> @aleadag Sorry for the late reply. I'll review and test the PR after v0.17.0.

Cool, feel free to let me know if there is anything I can help.

ganigeorgiev:
@aleadag When you find time, could you provide a short walkthrough what is required to create a WeChat OAuth2 app?

I'm having difficulties navigating through "Weixin Open Platform" with google translate in order to create my account to test the integration. There also seems to be some sort of verification fee and a lot of personal data are required for "Admin Info" that I definitely wouldn't provide. Could you confirm that this is the only way to register?

ganigeorgiev:
I've decided to close this PR as I'm not sure that I'll be to test the provider or even maintain it in the long run.
Adding the murky security and privacy concerns around the app and their parent company, I'm not sure it would be a good idea at this point to have it integrated in PocketBase.

Depending on the users demand I may reconsider it (_similar to the Apple OAuth2 integration_), but for now I'll close the PR.

Eventually in the future I'll also consider changing how the OAuth2 providers are managed, because due to the way how we store them at the moment as part of the typed `Settings` struct it is not very convenient and prevents extendability.

aleadag:
> I've decided to close this PR as I'm not sure that I'll be to test the provider or even maintain it in the long run. Adding the murky security and privacy concerns around the app and their parent company, I'm not sure it would be a good idea at this point to have it integrated in PocketBase.
> 

Understood your concern.

> Depending on the users demand I may reconsider it (_similar to the Apple OAuth2 integration_), but for now I'll close the PR.
> 
> Eventually in the future I'll also consider changing how the OAuth2 providers are managed, because due to the way how we store them at the moment as part of the typed `Settings` struct it is not very convenient and prevents extendability.

It sounds a good idea to move OAuth2 providers out of "Settings". OAuth2 providers would be just like "plugins". Before that, I will use PocketBase as a framework, and implement WeChat auth outside PocketBase.

Anyway, thanks for your time to look into this.

PR #2877, opened 2023-07-08
Always show error responseData.message on toast

svekko:
Simple use-case would be editing a record that doesn't exist anymore. 

To replicate: 
* Open 2 browser tabs. 
* Remove record in first tab, then try to edit that very same record in the second tab.

At the moment, from UI's point-of-view there is nothing indicating user that something went wrong. This PR fixes that.


ganigeorgiev:
Hm, I don't remember why I've done it this way. I'll have to think a little more on it but that will be after v0.17.0.

I'll close the PR for now as the suggested approach doesn't look correct - if we always show the error message the second 404 if check is confusing and may not be needed.

svekko:
> Hm, I don't remember why I've done it this way. I'll have to think a little more on it but that will be after v0.17.0.
> 
> I'll close the PR for now as the suggested approach doesn't look correct - if we always show the error message the second 404 if check is confusing and may not be needed.

Okay, that's cool. :)

The second 404 check was there to differentiate between 404 error due to not finding the API endpoint (toast will not be shown in that instance) and an 404 error returned by API handler (error will be shown in that instance).

ganigeorgiev:
I don't think the 404 check is needed at all, but I don't exactly remember why I've done it. It could be something forgotten from testing or similar. I'll think on it sometime later as I also have in my todo other similar things related to the deleted/missing records (eg. in the relations field items loader) that I plan to refactor.

ganigeorgiev:
I remembered why the 404 exclusion check is there - in its very early stage there was a dedicated 404 error page and the idea was that on 404 error the user would have been redirected to that (similar to 401/403 errors) but at some point I removed the error route and decided to use the index page as a "catch-all" route.

In other words, the 404 check shouldn't be needed anymore and I'll remove it with the next release.

Issue #2875, opened 2023-07-07
feature request: add response time in the request log section


Jeyanthinath:
Currently the field doesn't have the response time section

![image](https://github.com/pocketbase/pocketbase/assets/2253649/9c0e786b-19d7-42da-9289-d1bde02bbfcf)

This will be helpful in identifying the bottlenecks in the system.

ganigeorgiev:
`execTime` field was added to the refactored request logs data. The change will be available with the upcoming v0.20.0 release soon.

PR #2886, opened 2023-07-10
Added response time to logs in meta

WinCisky:
Added a timer to track the response time in the logs, the information is stored in the meta field. #2875


ganigeorgiev:
Thank you but this is not the way how I'm planning to implement it. The logs will be refactored in the near future as part of the ["logs generalization" task](https://github.com/orgs/pocketbase/projects/2/views/1?pane=issue&itemId=18223491) but for now remains a low priority.

PR #2957, opened 2023-07-20
improve http error handler

shynome:


shynome:
this is to support wrapped errors. echo support wrapped errors.

[`err2.Handle(&err)`](https://github.com/lainio/err2) auto wrap errors, if pocketbase support wrapped errors, I can remove the below hack code

```golang
	app.OnBeforeServe().Add(func(e *core.ServeEvent) error {
		eh := e.Router.HTTPErrorHandler
		e.Router.HTTPErrorHandler = func(c echo.Context, err error) {
			if v := new(apis.ApiError); errors.As(err, &v) {
				err = v
			} else if v := new(echo.HTTPError); errors.As(err, &v) {
				err = v
			}
			eh(c, err)
		}
		return nil
	})
```

shynome:
This is to support wrapped errors.

Why? `echo.DefaultHTTPErrorHandler` supports wrapped errors, this PR makes pocketbase have the same behavior as `echo`.

----
a better reason

ganigeorgiev:
Our case is a little more complicated as we have more than 1 type of errors.

I'll consider applying a similar check in the `develop` branch sometime later this week but will prioritize the `apis.ApiError`.

PR #3026, opened 2023-07-31
Fix URL in when in base path exists

sonyarianto:
Broken when we are in base path situation.
e.g.
Pocket installation is https://api.xxx.com/v2 then when go to oauth provider settings page will be broken

### DIFF
--- a/ui/src/components/collections/CollectionAuthOptionsTab.svelte
+++ b/ui/src/components/collections/CollectionAuthOptionsTab.svelte
@@ -181,7 +181,7 @@
         {#if collection.options.allowOAuth2Auth}
             <div class="block" transition:slide|local={{ duration: 150 }}>
                 <div class="flex p-t-base">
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
I believe the `href` should be just `/settings/auth-providers`. I'll update it manually sometime later today to avoid the back-and-forth.

ganigeorgiev:
I've merged the change in `master`.

Because the `use:link` action is not applied and the link has `target=_blank`, for this particular case I think the use of relative path as suggested in the PR is the correct one.
_In general the UI for this particular component may change in the future as currently opening  the OAuth2 providers page in a new tab is not very user friendly and may confuse the users._

The merged fix will be available with the next minor v0.17.1 release sometime later this week.

sonyarianto:
@ganigeorgiev nice thank you Gani

sonyarianto:
agree, better not using open new tab, personally always afraid if a web doing open new tab hahaha

PR #3192, opened 2023-08-25
fix: restore previous api rules when downgrading migration

impact-merlinzerbe:
When changing api rules, the generated `down` functions clear the api rules instead of setting them to their previous value. So when starting from state A and performing a change to an api rule that is not null/nil/admin-only and subsequently running a `migrate down` one ends up in state B which is missing the api rule present in state A.

This PR changes the code generation such that the `down` functions set api rules to their previous value instead of clearing them.

### DIFF
--- a/plugins/migratecmd/migratecmd_test.go
+++ b/plugins/migratecmd/migratecmd_test.go
@@ -381,7 +381,7 @@ migrate((db) => {
 
   collection.name = "test456_update"
   collection.type = "base"
-  collection.listRule = null
+  collection.listRule = "@request.auth.id != ''"
   collection.deleteRule = "updated > 0 && @request.auth.id != ''"
   collection.options = {}
   collection.indexes = [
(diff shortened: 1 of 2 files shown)


plugins/migratecmd/migratecmd_test.go
```diff
@@ -779,3 +779,150 @@ func TestAutomigrateCollectionNoChanges(t *testing.T) {
 		}
 	}
 }
+
+func TestAutomigrateRuleDowngrade(t *testing.T) {
```
ganigeorgiev, line 783: To avoid maintaining another test, could you update the existing `TestAutomigrateCollectionUpdate`?

ganigeorgiev (changes requested):
Thanks, looks a lot better this way, I don't know why I've implemented it with the messy if/elseif checks.

_I left one minor note about the test. Once done, I think it can be merged._

plugins/migratecmd/migratecmd_test.go
impact-merlinmarek: done

ganigeorgiev:
Thanks! To avoid the back-and-forth I'll merge it with the latest pushed change and will add update only the test to include also the other rule fields for slightly better coverage.

PR #3327, opened 2023-09-16
Feat/add jsonschema json option

### LINKED ISSUE #1154: Feature Request: JSON Schema to Field Type JSON
Hi, Thanks for such a wonderful software.
It would be nice to have ability to either
a) add link to JSON schema URL
or
b) paste the schema directly in database
or
c) upload json schema file
while adding/editing collection

we can then use schema to 
primarily: send schema with api requests
optionally: validate json entry with schema in PB itself.

Thanks!

ItayBenAmi:
Adds the option to validate JSON fields with a given JSON schema option. 

This is done by adding JsonSchema as an optional string field in `schema.JsonOptions` and validating the JSON value with the JSON schema in the `validators.checkJsonValue` check. 

The JSON schema validation was written using the https://pkg.go.dev/github.com/santhosh-tekuri/jsonschema/v5 library, which leads the line in terms of performance and correctness of GO JSON schema validator libraries, as can be seen here: [https://dev.to/vearutop/benchmarking-correctness-and-performance-of-go-json-schema-validators-3247](url). It also supports the latest meta-schema (Draft 2020-12), which is used in the `checkJsonSchema` function to validate a given JSON schema.

Closes #1154.


ganigeorgiev:
Thank you, but I'm not sure about the suggested implementation and I'll close the PR for now.

The linked task remains with a very low priority (_see also https://github.com/pocketbase/pocketbase/discussions/3271_).

Couple things that I wish eventually to tackle if this ended up getting merged:

- The Admin UI `json` field options also needs to be updated to reflect the change.

- Check if possible to skip the external dependencies, even if that means implementing only  the latest version of the spec (_I haven't checked the JSON spec to see how much work that will require and whether it will be feasible at all_).

ItayBenAmi:
Thank you very much for your time.

As for the two things you mentioned:

- I added the option in the admin ui in my local branch and can open a new PR if necessary.
- I believe that implementing JSON schema validation in line with the latest spec without any external dependencies would
  overly complicate things so I would recommend using an external library if you decide to implement this feature in the future.

PR #3374, opened 2023-09-23
Make `@request.query` and `@request.headers` fields work on realtime rules

### LINKED ISSUE #2514: Feature Request: `@request.query.*` for realtime API
Hi,

Would it be possible to pass some extra query parameters to realtime API - in a similar way how it's currently possible to do it for simple records API - to be used in listRule?

Right now, I can create a collection "messages" with a field "key", and add a list API rule like this: `@request.query.key = key` - this way anonymous users can find all messages if they know a "key" for them, like this: `pb.collection('messages').getFullList({key:"..."})`.

However, this currently doesn't work for realtime API - it doesn't have a "query" option, so `@request.query.key` is always empty for them. In other words, realtime API works only for those records which you can see by issuing a simple `pb.collection('messages').getFullList()` call.

I was able to workaround it by creating a user for anyone accessing such "key-protected" records, recording all known keys into this user's "known" multi-relation field, and adding a listRule like this: `@request.auth.known.key?=key`. But I wonder if you might consider implementing it in the same way as it's done for records API. Thanks!

adario7:
The [docs](https://pocketbase.io/docs/api-realtime/) state that

> When you subscribe to an entire collection, the collection's ListRule will be used to determine whether the subscriber has access to receive the event message.

however headers and query fields are not currently being populated, making ListRules behave differently on realtime queries.
should fix #2514.

note: the SDKs don't support query and headers in subscriptions yet, so to test the feature i used a custom http client to inject headers/query params:

```dart
class MyClient extends http.BaseClient {
  final http.Client _httpClient = http.Client();
  MyClient();
  @override
  Future<http.StreamedResponse> send(http.BaseRequest r0) {
    const inject = <String, String>{
      "foo": "bar"
    };
    var request = http.Request(
      r0.method,
      r0.url.replace(queryParameters: {...r0.url.queryParameters, ...inject}),
    );
    if (r0 is http.Request) {
      request.body = r0.body;      
      request.encoding = r0.encoding;      
    }
    request.headers.addAll({ ...r0.headers, ...inject });
    return _httpClient.send(request);
  }
}
```

### DIFF
--- a/apis/middlewares.go
+++ b/apis/middlewares.go
@@ -24,6 +24,8 @@ const (
 	ContextAdminKey      string = "admin"
 	ContextAuthRecordKey string = "authRecord"
 	ContextCollectionKey string = "collection"
+	ContextQueryKey      string = "query"
+	ContextHeadersKey    string = "headers"
 )
 
 // RequireGuestOnly middleware requires a request to NOT have a valid
--- a/apis/realtime.go
+++ b/apis/realtime.go
@@ -155,6 +155,18 @@ func (api *realtimeApi) connect(c echo.Context) error {
 	}
 }
 
+// extract the first value of all entries and normalizes the keys
+// ("X-Token" is converted to "x_token")
+func normalizeMap(input map[string][]string) map[string]any {
+	res := map[string]any{}
+	for k, v := range input {
+		if len(v) > 0 {
+			res[strings.ToLower(strings.ReplaceAll(k, "-", "_"))] = v[0]
+		}
+	}
+	return res
+}
+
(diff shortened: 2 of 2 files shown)


ganigeorgiev:
Thank you but I don't think the proposed implementation is OK.

If we ended up adding support for realtime headers and query parameters it must be per subscription topic not "globally" for all client subscriptions.

The main problem with this feature request is actually on the client-side because I'm not sure at the moment how to implement it without breaking the existing SDKs APIs.

But for now this feature request remains with a low priority (_see also https://github.com/pocketbase/pocketbase/discussions/3271_).

PR #3430, opened 2023-10-01
refactor: replace interface{} with any

KunalSin9h:


ganigeorgiev:
Thank you but I don't think this is really necessary and it shouldn't have any effect for the end user.

I'll consider updating in the `develop` branch the 2 Go files sometime before v0.19 but it is definitely not something important.

Also note that the `types.d.ts` are auto generated and the file shouldn't be edited by hand.

KunalSin9h:
@ganigeorgiev ok

PR #3533, opened 2023-10-16
feat: ignore workflow run when only *.md file changes

KunalSin9h:
The most recent workflow runs are only triggered due to change in Readme.md file, if i am not mistaken their is a limit (quota) for workflow run time for organization, and these workflow run do not give any benefits, so we can ignore workflow runs when only `*.md` files are changed. 

This PR change the workflow file to ignore the workflow run when only `*.md` files like `README.md`, `CHANGELOG.md`, `CONTRIBUTING.md`, `LICENSE.md` and `README.md` are changed.

image

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,11 @@ name: basebuild
 
 on:
   pull_request:
+    paths-ignore:
+      - "*.md"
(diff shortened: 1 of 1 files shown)


ganigeorgiev:

PR #3591, opened 2023-10-23
Display name in field schema, #3590

koltyakov:
The PR implements an extra metadata field for storing display name.

This is a draft PR. If @ganigeorgiev have no principal disagreement with adding such a feature, I can add tests and clean ups, etc.


ganigeorgiev:
Thank you but as mentioned in the related discussion, this remains out of the scope for now.

It is not the specific implementation, nor the effort required for it that I'm against adding it in PocketBase for now.

koltyakov:
OK

PR #3602, opened 2023-10-25
use collection name in oauth docs

j0sh:


ganigeorgiev:
Thanks! I've pushed it manually in the `develop` branch as `ui/dist` needs also to be generated.

PR #3689, opened 2023-11-05
Relax OAuth2 RedirectUrl validation

sergeypdev:
Hey, first of all thanks for making PocketBase! 

I encountered an issue with OAuth2 when implementing auth in iOS. Current RedirectUrl validation is too restrictive and doesn't allow custom url schemes, which is needed for native apps.

I think it's best to disable that validation on the PocketBase side, because it's not really necessary there. Let the OAuth provider handle that instead in their own way.

### DIFF
--- a/forms/record_oauth2_login.go
+++ b/forms/record_oauth2_login.go
@@ -7,10 +7,8 @@ import (
 	"time"
 
 	validation "github.com/go-ozzo/ozzo-validation/v4"
-	"github.com/go-ozzo/ozzo-validation/v4/is"
 	"github.com/pocketbase/pocketbase/core"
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Could you also add a test with a custom url scheme in the `record_oauth2_login_test.go`?

_Side-note: when using the ["all-in-one" OAuth2 flow](https://pocketbase.io/docs/authentication/#oauth2-integration) you don't need a deeplink._

ganigeorgiev:
I've squash merged the changes in `master` since I wanted to make it part of the v0.19.3 release that will be available sometime later today.

sergeypdev:
@ganigeorgiev thank you!

Issue #3759, opened 2023-11-15
High memory usage when resizing thumbnails


t-muehlberger:
Hello, thank you for your great work on PocketBase, I used it for a little internal project and was pretty happy with how it went. I only have one issue. The process uses a lot of memory (over 1.5GB) when resizing thumbnails and eventually is OOM-Killed.

My tool lets people upload high resolution photos (many are over 5MB). When I open that PB-Dashboard to view the collection, the browser tries to load many thumbnails (40 per page) at the same time. This causes the server to use a lot of memory and eventually crash (usually at the second Page).  This behavior can only be observed on newly uploaded photos for which the thumbnails have not yet been generated. 

Now I understand that resizing large images requires that they are kept in memory and some memory usage is inevitable. But I am running the container with a 2GB memory limit and I think this should be sufficient. 

Also what seems a bit strange to me is that the memory usage is way above the actual size of the images. Resizing 80 Images with around 5MB each (400MB total) results in a memory consumption of over 2GB. 

**Possible Solution:** I think it would be really important to be able to limit the number of thumbnails that are generated concurrently. 

Keep in mind, that generating many many thumbnails at the same time will probably slow the server down causing the next thumbnails to be generated even slower and thus kept in memory even longer.

Also might be the case that the same thumbnail is generated more than once but I am not sure about that. I can Imagine that users refreshing the page when thumbnails are slow to load makes the problem even worse because the browser might be requesting the same thumbnail, that is already processing, again. 

It might also be possible that the concurrency is not the core issue here and that there is a memory leak somewhere but I don't know about that. Definitely should be further investigated, it is possible that limiting the concurrency would only hide some of the symptoms of a different problem.

Also, if you can point me into the right direction I can try to create a PR myself to fix the issue.

ganigeorgiev:
If you are experiencing memory issues, you can try defining the [`GOMEMLIMIT=500MiB` env variable](https://pkg.go.dev/runtime) (_it is a soft limit and will instruct the GC to run more often once the threshold is reached_).

I don't think that there is a memory leak, but bulks thumb generation can definitely be improved.

We can try to create a worker pool and restrict the thumb generation processes. I'm currently busy with other tasks, but if anyone want to pick it the thumbs generation code is located in 
https://github.com/pocketbase/pocketbase/blob/985ab1e5b79a328b516decca3d1d7dc8edfcbba0/tools/filesystem/filesystem.go#L393.
[...]

Issue #2231, opened 2023-04-06
S3 file storage not working with Google cloud storage since 0.14.1


joelml:
Hi! First of all, thanks for amazing work!

I've been using the s3 storage to connect to a bucket in GCS without a problem up until the latest changes to the AWS SDK.
PR: #1346 

I've been trying to figure out what could be wrong but I'm not skilled in Go unfortunately.
I'm getting an error (403) that the test file couldn't be uploaded and it doesn't matter if I'm upgrading from 0.14.0 or starting fresh at latest 0.14.2.
But using the same credentials for the s3 storage works fine going back to 0.14.0. (also using the old database).

Any ideas?

ganigeorgiev:
Could you start the executable with `--debug` and check if there is a more detailed error message in the console?

ganigeorgiev:
Hm, I think this could be related to https://github.com/aws/aws-sdk-go-v2/issues/1816, which means that it will require some special handling for GCS due to the changes of how v2 sign and send the request headers.

I'll consider a revert sometime later today.

joelml:
Thank you for looking into this!

Using `--debug` only prints out the params being set as normal, nothing else.

ganigeorgiev:
To avoid having a special implementation just for GCS, the AWS SDK change was reverted in [v0.14.3](https://github.com/pocketbase/pocketbase/releases/tag/v0.14.3) (we don't use any of the newer S3 apis anyway).
If you are still having difficulties with the S3 setup, feel free to let me know and I'll investigate it further.
_Note that `--debug` will show additional error messages in the console only during an action (aka. if you want to check the error with the v2 of the sdk you have to trigger S3 settings update)._

PR #3794, opened 2023-11-20
WIP: Limit Thumbnail Generation Concurrency

t-muehlberger:
This should fix the issue described in #3759. I tested this solution in my PB deployment and it did not run out of memory anymore. 

This solution does not use a separate Worker-Pool, rather it just blocks the request handler if too many thumbnails are generated at the same time. I think this is the simpler approach. 

It will also be easier to remove the semaphore in case this fix would be replaced by a more general solution in the new version FS abstraction (like limiting the number of open files or the number of bytes stored in memory). 

The limit should probably be configurable. @ganigeorgiev do you think I should read the Env-Variable `MAX_THUMBNAIL_CONCURRENCY`?

### DIFF
--- a/apis/file.go
+++ b/apis/file.go
@@ -1,11 +1,13 @@
 package apis
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/labstack/echo/v5"
 	"github.com/pocketbase/pocketbase/core"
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
Without tests to show that the thumbs generation is the culprit and not the `aws-sdk-go` file upload I don't feel confident merging it, sorry.

Based on what `runtime.NumCPU() * 2` is decided? Does it have significant impact on the memory if it is higher? Could you elaborate how you are testing to try to reproduce your results?
What is your memory usage when use `GOMEMLIMIT` set to ~70% of your max memory?

Some notes on the code:
- The semaphore variable could be moved as field inside the `fileApi` struct.
- I'm not sure about whether it would be better to hold the connection open instead of returning 503 (or even returning the original image instead).
- The double `fs.Exists` ideally should be avoided if possible. In the pasts users has complained about hitting S3 rate limits (see failed https://github.com/pocketbase/pocketbase/issues/28) and I want to minimize the extra calls if possible (_eg. we can store the current generating thumb name in a concurrent `map[string]struct{}{}`_).

Note that it is possible the higher memory consumption to be caused by https://github.com/aws/aws-sdk-go/issues/4916.

Before complicating further the implementation I suggest to prepare a test case (or at least some manual steps) that consistently trigger the issue and to start profile from there.
We can try changing back to `aws-sdk-go-v2` to see if there would be any improvement (we've used v2 in the past but had to revert to v1 because of https://github.com/pocketbase/pocketbase/issues/2231). If v2 perform better we can try to upgrade again and apply a fix for the GCS headers and avoid the locks altogether.

t-muehlberger:
Alright, let me get back to you with a benchmark or script to replicate the issue.

`runtime.NumCPU() * 2` seemed to utilize the CPU pretty well for my small Hetzner VPS with Backblaze B2 as storage bucket.  This would assume that 50% of time would be spent fetching images (IO-Bound) and 50% resizing images (CPU-Bound) but I did not measure that. Also the factors involved (S3 Latency, Image Size, CPU/RAM available) would vary widely between different deployments. 

With limited concurrency the memory usage always stayed around `GOMEMLIMIT` or a little bit above, even when setting `GOMEMLIMIT` as low as 256MiB. The memory that is actually in use at any moment is probably a lot less, with a lot of garbage being generated and cleaned as soon as `GOMEMLIMIT` is reached. 

I will keep digging deeper into the issue, but it is probably going to take me a few days to find the time. You cen close the PR in the mean time if you want to keep github clean.

ganigeorgiev:
I was able to experiment a little with this and can confirm that the the thumb generation or large images requires a lot of memory allocations and the the problem is not just the S3 file upload as the higher memory consumption can be examine even when only the local filesystem is used.

I've performed a test locally with ~7MB 9319x5792px jpg and generating 3 different thumbs concurrently for both storage options - 100x100, 500x500 and 1000x1000. `runtime.MemStats.Alloc` indicated ~1.2GB for the local filesystem case and ~2GB for the S3. Encoding the images just with the std `image/jpeg` also seems to produce similar results which I guess make sense as `imaging` uses the std lib under the hood (in all cases `draw.NearestNeighbor` resampling filter was used as a base).

So I don't think there is much we can do (at least not easily) in this case for optimizing the memory allocations.

I've pushed a slightly modified version of your PR in the `develop` branch that also should cover the case where multiple concurrent requests for the same thumb are being executed (there is also a small automated test to ensure that new thumbs are being saved properly). The changes will be available with the next v0.20.0 release soon.

I still don't exactly like the `runtime.NumCPU()` usage (by default I changed it to `runtime.NumCPU() + 1`) because the number of logical CPUs are not directly related to the memory allocations, which is the actual problem here, but since I don't have a better solution at the moment I'll keep it for now as it is and eventually will adjust it in the future based on the users feedback and my tests with Presentator.

But in general, if you are still experiencing OOM errors, you can try combining the latest changes with `GOMEMLIMIT` as this is expected to keep the GC allocations within your desired limit (_I'll consider adding a note in the documentation for it_).

PR #3948, opened 2023-12-13
Add Bitbucket OAuth provider

aabajyan:
Hello, this is my first commit. If there's any issues let me know!

I tested with JS-SDK and I managed to sign in.
I ran `make test` and it seems to work fine.
I also ran `make lint` and no new errors was found.

You can create a new OAuth app here https://support.atlassian.com/bitbucket-cloud/docs/use-oauth-on-bitbucket-cloud/


tools/auth/bitbucket.go
```diff
@@ -0,0 +1,115 @@
...
+	*baseProvider
+}
+
+// NewBitbucketProvider creates a new Bitbucket provider instance with some defaults.
+func NewBitbucketProvider() *Bitbucket {
+	return &Bitbucket{&baseProvider{
+		ctx:         context.Background(),
+		displayName: "Bitbucket",
+		pkce:        false,
+		scopes:      []string{"account"},
```
ganigeorgiev, line 29: Do we need also the [`email`](https://developer.atlassian.com/cloud/bitbucket/rest/intro/#email) or `account` includes it?

ganigeorgiev (commented):
Thank you! I left some comments, feel free to check them when you find time.

I'll be able to test and review the integration in more details over the weekend.

tools/auth/bitbucket.go
```diff
@@ -0,0 +1,115 @@
...
+
+	rawUser := map[string]any{}
+	if err := json.Unmarshal(data, &rawUser); err != nil {
+		return nil, err
+	}
+
+	extracted := struct {
+		UUID     string `json:"uuid"`
+		Username string `json:"username"`
+		Nickname string `json:"nickname"`
```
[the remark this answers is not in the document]
aabajyan: Yeah it does exists, but you are right I should use the listed one since it's the same

tools/auth/bitbucket.go
aabajyan: account scope includes email too yes

ganigeorgiev:
I was able to test the integration and everything seems to work fine!

I've squash-merged the PR locally in the `develop` branch with some minor modifications:
  - added checks for the `account_status:active` field in case the account is suspended for some reason
  - refactored the primary email function to ignore only the request errors for consistency with the other similar providers and to help with the debugging in case something else has gone wrong

The provider will be part of the next v0.21.0 release (_there is no ETA yet_).

PR #4035, opened 2023-12-26
nit: replacing redundant 'JWT token' with 'JWT'

calvin-barker:
"JWT" stands for "JSON web token" (https://datatracker.ietf.org/doc/html/rfc7519), so "JWT token" would mean "JSON web token token." This is a minor pet peeve of mine, so I feel it is my duty to fix this.


ganigeorgiev:
Thanks. I've squash merged it locally in the `develop` branch to avoid the back-and-forth as the autogenerated types files shouldn't be updated by hand.

Issue #4066, opened 2024-01-02
The act of resetting a password should automatically mark a user as "verified"


burggraf:
Scenario:  The user signs up and never receives a confirmation email.  The user clicks on "forgot password" and gets a password reset link, click it, and changes their password.  At this point they are still "unverified", though I would argue that ARE verified since they received an email at the given address and clicked on it to verify the address is valid.

This would be the solution to "I didn't get a verification email" -- answer: "just reset your password."

ganigeorgiev:
It makes sense, I'll consider it for the next release.

burggraf:
I figured I could hack it with:
`onRecordAfterConfirmPasswordResetRequest`
but that doesn't seem to be firing when I click to reset my password, or maybe I'm misunderstanding that hook?

ganigeorgiev:
The `onRecordAfterConfirmPasswordResetRequest` will be fired **after** the user clicks on the email link and successfully resets their password.

To avoid resaving the record, you can use its before hook, something like:

```js
onRecordBeforeConfirmPasswordResetRequest((e) => {
    if (e.record) {
        e.record.setVerified(true);
    }
})
```

burggraf:
I just never see that trigger firing for some reason.

burggraf:
Never mind, my mistake.  Environment isn't set up right.

burggraf:
Yep, that works.  Hooks rock.

ganigeorgiev:
This is implemented in the `develop` branch and will be available with the next v0.21.0 release (_there are no ETAs yet_).

PR #4103, opened 2024-01-07
Add Option to Verify users on pwd reset

SparshKaushik:
Hey,
This Pull Request adds an option in auth Collections to allow users to automatically get verified after doing a successfull password reset #4066.

UI :-
image

Backend :-
Added a option just like MinPasswordLength and a check to mark user as verified when the option is enable (i.e. true)

Default Value :-
Enabled (true)

Tests :-
- UI (Default auth collection, Creating new auth collection, pwd reset with enabled and disabled)
- Go Unit tests (Honestly, I have no idea what i did in there)

This is my first ever OpenSource Contribution, so Please Correct me if I am wrong anywhere.
Hopefully will be making more pull requests here soon :tm:. Lets Get Pocketbase to v1.0.0 :confetti_ball:


ganigeorgiev:
Thank you but I don't think there is a need for an option and we can make it a default behavior without extra knobs. We'll also need tests for the confirm API endpoint.

In any case this is in the roadmap and will be implement with the next release but there are no ETAs.

_Regarding features contribution, please check the note in https://github.com/pocketbase/pocketbase?tab=readme-ov-file#contributing._

PR #4204, opened 2024-01-21
Fix typo in "unverified"

tylerlaprade:
`s/unverfied/unverified`


ganigeorgiev:
Thank you. I'll apply the change manually in the `develop` branch as the `ui/dist` need to be rebuild.

tylerlaprade:
Thanks @ganigeorgiev. For next time, how do I rebuild it myself? I saw that file when searching but assumed it had to be built, not manually modified, so I didn't touch it.

PR #4309, opened 2024-02-07
Update README.md with additional command necessary for installation 

client4:
When installing pocketbase with go version 'go1.22.0 linux/amd64' you get an error about not being able to find packages.  'go mod init xxxxx' is required before the go get command. 

I chose pocketbase as the default 'go mod init' but it may make sense to let the reader know they could use something else, like 'example.com/pocketbase' or 'go mod init $MODULENAME'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ your own custom app specific business logic and still have a single portable exe
 
 ```sh
 # go 1.21+
+go mod init pocketbase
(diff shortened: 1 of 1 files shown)


ganigeorgiev:

Actually the "Installation" section can be removed and replaced with the content from https://pocketbase.io/docs/go-overview/ (I'll consider updating it sometime later).

PR #4486, opened 2024-03-04
feat: add new api to handle store address

ternakkode:


ganigeorgiev:
Sorry but this doesn't make sense. Also please check  https://github.com/pocketbase/pocketbase/discussions/4355.

PR #4490, opened 2024-03-05
Add code for taking a json payload, validating it and creating a reco…

bhabyajain02:
A custom endpoint /validatePayloadAndAddRecord is exposed which takes a json payload in the request. 
It does validation on the json payload. The validations are -
1. Password must be present.
2. PasswordConfirm must match with Password.

If the validation is passed, it adds the record to users collection.

The testing was done and here are the snapshots for the same.
Case 1: Password is not specified -
image

Case 2: Password and PasswordConfirm did not match -
image

Case 3: Validation passed and record added -
image

image

More tests like adding multiple records, adding same record again, etc were performed and got the expected result.

### DIFF
--- a/sample/main.go
+++ b/sample/main.go
@@ -1,0 +1,104 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/url"
+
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
I don't understand what is the purpose of the above and I'd assume that you've opened a PR to a wrong repo.

PR #4505, opened 2024-03-06
Fix code-wrapper height for Firefox

seriousm4x:
When generating a log event as follows:

```go
app.Logger().WithGroup("update").Error("some error message", "arg1", "info here")
```

The line height in the admin UI appears too small in Firefox.

On the left is Firefox 123.0.1, and on the right is Chromium 122.0.6261.94:
swappy-20240306_155535

Removing `max-height: 100%;` from the `code-wrapper` class will resolve this issue.

(Sorry, just minutes late for the 0.22.3 release 😄)

### DIFF
--- a/ui/src/components/base/CodeBlock.svelte
+++ b/ui/src/components/base/CodeBlock.svelte
@@ -41,7 +41,6 @@
     .code-wrapper {
         display: block;
         width: 100%;
-        max-height: 100%;
         overflow: auto; /* fallback */
         overflow: overlay;
     }


ganigeorgiev:
I don't exactly remember I've added the `max-height`. Based on the other styles I think it was probably because the code block needs to have a scroll if the parent container has fixed height.

I'll investigate it a little bit later today and eventually this may require adding a special case only when used inside a table (_I'm not sure why Firefox renders it this way in the first place_).

ganigeorgiev:
Based on the git history the max-height and its accompanying overflow styles seem to have been added because of the old collections Export/Import UI, but it is no longer needed since that par has changed so it should be safe to be removed.

I've pushed it as a separate commit in `master` since the `ui/dist` also needs to be updated but I'll not rush another release as it is not something critical.

PR #4575, opened 2024-03-16
Auth Provider For LinkedIn

vitalii-bulyzhyn:
Hey PacketBase team!

Really enjoyed product you've built, but I noticed that there were no LinkedIn auth provider which I needed for my app, so I adding it in this PR. Hence I tried to follow your pattern, it's my first PR for PocketBase and I would love any feedback on it


ganigeorgiev:
Thank you but I don't there is need for this as LinkedIn already has support for OpenID Connect (OIDC) and you can use the generic OIDC provider.
Just make sure to disable the "PKCE" option from the settings because LinkedIn doesn't support it based on the users reports - https://github.com/pocketbase/pocketbase/discussions/3799#discussioncomment-7828549.

PR #4672, opened 2024-03-31
Add Svelte Inspector Toggle

moalamri:
By default, the inspector toggle button will only be visible in dev env, as it's always set to `false` in build.
The toggler is useful to easily jump to component level source while developing.


ganigeorgiev:
Thank you but I don't think this is necessary at the moment.

Issue #351, opened 2022-08-25
Mime type checking does not work for .wav files


alexmercier25:
Mime types permitted:
audio/x-wav,audio/mpeg

but still throws an error:
2022/08/25 16:56:16 audio_file: The following mime types are only allowed: audio/x-wav,audio/mpeg..

However it works great with mp3 files.

ganigeorgiev:
PocketBase uses `http.DetectContentType()` from the standard Go library, which has some limitations and quirks.

For full list with the supported mimetypes and their signatures you could check - https://github.com/golang/go/blob/8b364451e2e2f2f816ed877a4639d9342279f299/src/net/http/sniff.go#L66-L195.

There is no `audio/x-wav` in the above list, BUT there is `audio/wav` so you could try replacing it with that.

ganigeorgiev:
For clarity, here is the full list with the supported mime types:

```
text/html; charset=utf-8
text/xml; charset=utf-8
text/plain; charset=utf-16be
text/plain; charset=utf-16le
text/plain; charset=utf-8
application/pdf
application/postscript
image/jpeg
image/png
image/gif
image/bmp
image/webp
image/x-icon
audio/aiff
audio/mpeg
audio/midi
audio/wave
application/ogg
video/avi
video/mp4
video/webm
font/ttf
font/otf
font/collection
font/woff
font/woff2
application/vnd.ms-fontobject
application/x-gzip
application/zip
application/x-rar-compressed
application/x-rar-compressed
application/wasm
---
# everything else will be detected as:
application/octet-stream
```

I'll update the "Mime type" field validator in the next v0.6.0 release to be more "strict" and allow only the above mime types.

ganigeorgiev:
Instead of restricting the supported mimetypes, and for better distinguishing document files from `application/zip`, I've decided to replace the standard library with [gabriel-vasile/mimetype](https://github.com/gabriel-vasile/mimetype), which has far more complete list of file singatures - https://github.com/gabriel-vasile/mimetype/blob/master/supported_mimes.md
_(when serving files the standard go lib still will be used because `gocloud/blob` content-type check is based on it, but that's ok since the types we are serving inline are already supported)_.

This is currently only in the `develop` branch, but will be publicly available with the next v0.6.0 release (_sometime next week or the week following it_).

[...]

PR #4682, opened 2024-04-01
Added MIME Type Support as well as caching support for StaticDirector…

theHamdiz:
Added MIME Type Support as well as caching support for StaticDirectoryHandler as well as depended on std mime package instead of github.com/gabriel-vasile/mimetype which might if employed globally reduce external dependencies.

Also added caching support for the same handler.

### DIFF
--- a/apis/base.go
+++ b/apis/base.go
@@ -7,8 +7,10 @@ import (
 	"fmt"
 	"io/fs"
 	"log/slog"
+	"mime"
 	"net/http"
 	"net/url"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Sorry but I don't think there is need for this.

Using `gabriel-vasile/mimetype` package was an early decision due to limitations of the std library - https://github.com/pocketbase/pocketbase/issues/351.

Also caching doesn't need to be "hardcoded" into the handler and it can be implemented as middleware.

Also keep in mind that PocketBase is currently in the process of heavy internal refactoring as part of https://github.com/pocketbase/pocketbase/discussions/4355 and PRs for new features will not be accepted.

theHamdiz:
Thanks for illustration, I just wanted to participate in a project a believe in that's all.

PR #4711, opened 2024-04-06
feat: allow the admin ui to be "detached"

hugorut:
This change allows users to customize if the admin UI is embeded at build time or specified as a detached directory which exists outside the binary. This allows users to provide a customized admin UI provided at runtime. Additionally it offers the ability to reduce the binary size for users who do not care about running the admin UI in production.

The conditional embedding support is provided through build tags. By default the admin UI is embeded, users can specify a detached admin UI with by building the project with the `detachableadmin` build tag and specifying the directory for their detached admin UI in the `PB_ADMIN_UI_DIR` env variable.

### DIFF
--- a/ui/detached.go
+++ b/ui/detached.go
@@ -1,0 +1,13 @@
+//go:build detachableadmin
+
+package ui
+
+import (
+	"os"
+
+	"github.com/labstack/echo/v5"
+)
+
+func init() {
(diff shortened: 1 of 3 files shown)


ganigeorgiev:
Sorry but I don't think this is needed. If you want to embed your own Admin UI you can simply register a new endpoint and use the [StaticDirectoryHandler](https://pocketbase.io/docs/go-routing/#serving-static-files).

Additionally please check https://github.com/pocketbase/pocketbase/discussions/4355.

PR #4727, opened 2024-04-10
Add s390x and ppc64le support

lysliu:
Add s390x and ppc64le support


ganigeorgiev:
Thank you but let's not add this for now as I'm not sure if it is worth the release builds slow down since these platforms are not that commonly used.

I'll reconsider it after the [refactoring](https://github.com/pocketbase/pocketbase/discussions/4355).

PR #4957, opened 2024-05-19
removed redundant rand.Seed

xnpltn:
rand.Seed is deprecated, no reason to call the function in newer versions of go...


ganigeorgiev:
Thank you. I'll consider removing it with the [refactoring](https://github.com/pocketbase/pocketbase/discussions/4355) but for now there is no need for this change.

PR #4988, opened 2024-05-25
Adding one-click deploy button for RepoCloud.io to README.md

cosark:
Adding button enabling community access to one-click deploy template for self-hosting on RepoCloud.io


ganigeorgiev:
Thank you but no.

PR #5179, opened 2024-07-06
Fix days calculation bug for the old logs

nehmeroumani:
Corrected daysDiff calculation by dividing hours by 24 instead of multiplying.


sanbei101 (commented):
Indeed it is

ganigeorgiev:
_Just for info in case someone is following here - this change actually introduced a side-effect and due to the way how the checks inside the condition are performed it will actually keep the logs for 2x the specified period. An updated version that doesn't rely on the logs batch write will be available with the next minor release sometime later this week._

PR #5233, opened 2024-07-18
Added the possibility to make collection view and collection list requests visible for any authorized user

CraftException:
## Motivation

Hi there,
we are currently working on a relatively big SaaS with many collections and validations (PocketBase works great for that).
We want to keep the frontend interactive (currently a Vue.js SPA) and therefore wants to provide direct error hints for the user before sending the request to the server and letting it validates the request and sending error mesages back. But that would mean that we have to always manually keep the client validation in sync with the of the server (otherwise the request could be blocked from the server, although it is valid for the client or viceversa). That is sup-optimal for our workflow so we decided that it would be  great that any user can see the collection schemas and their validation rules

## This Pull Request

That's what we did here: Allowing the administrator to choose in the Application Settings to decide whether the collection view (GET /api/collections/:collection) and list (GET /api/collections) should only be visible for administrators, or for any authorized User. We created for that another property in the MetaConfig called `PublicSchemas` and the `RequireCollectionListAuth` Middleware in the `middleware.go` file, which only allows the request if either the PublicSchemas property is set to `true` and the user is anyhow authorized or if the property is set to `false` and the user is an authorized administrator.

## Is this a security problem?

We don't think so, because many fields in the schema (e.g. minlength, maxlength), are exposed on invalid requests anyway. You can say the Authorization Rules are none of the users's business, but for our SPA, this was a huge feature, because the page can now automatically hide fields and parts of the side that the user wouldn't be able to manage anyway and we didn't have to manually implement these rules on your client-side. Nevertheless, this is just an optional feature, if you don't need/ don't want it, just disable it.


ganigeorgiev:
I'm sorry but there are no plans for this. Please check the contribution note in the readme and https://github.com/pocketbase/pocketbase/discussions/4355.

PR #5283, opened 2024-07-31
allow server graceful shutdown using context cancellation

telemac:
Hello, thank you for your great work on pocketbase.

I am integration pocketbase as a library in an application that already handles ctrl-c and interrupt signals.

This is my attempt to make pocketbase interruptible by context cancellation without modifying the api and the actual behavior.

I did not find another way to do this without modifying pocketbase, if I missed something let me know.

Here is an usage sample, if ServerCtx is omitted we keep the original behavior :

```go
package main

import (
	"context"
	"github.com/pocketbase/pocketbase/apis"
	"log/slog"
	"time"
)

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
	defer cancel()

	app := pocketbase.New()
	app.Bootstrap()

	_, err := apis.Serve(app, apis.ServeConfig{
		HttpAddr:        "0.0.0.0:8090",
		ShowStartBanner: true,
		AllowedOrigins:  []string{"*"},
		ServerCtx:       ctx,
	})
	slog.Info("apis.Serve stopped", "error", err)

}

```

### DIFF
--- a/apis/serve.go
+++ b/apis/serve.go
@@ -45,6 +45,9 @@ type ServeConfig struct {
 
 	// AllowedOrigins is an optional list of CORS origins (default to "*").
 	AllowedOrigins []string
+
+	// ServerCtx is an optional context to be used by the server.
+	ServerCtx context.Context
 }
 
 // Serve starts a new app web server.
@@ -142,7 +145,11 @@ func Serve(app core.App, config ServeConfig) (*http.Server, error) {
 
 	// base request context used for cancelling long running requests
 	// like the SSE connections
-	baseCtx, cancelBaseCtx := context.WithCancel(context.Background())
+	ctx := context.Background()
+	if config.ServerCtx != nil {
+		ctx = config.ServerCtx
+	}
+	baseCtx, cancelBaseCtx := context.WithCancel(ctx)
 	defer cancelBaseCtx()
 
 	server := &http.Server{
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
I don't think this is needed. You can attach to the [`OnBeforeServe`](https://pocketbase.io/docs/go-event-hooks/#onbeforeserve) and modify the `e.Server` field.

Note that we already handle graceful shutdown based on the server context cancellation and an alternative approach to "kill" the PocketBase server (and to notify other parts of PocketBase) is to trigger the `Terminate` hook:

```go
app.OnTerminate().Trigger(&TerminateEvent{ App: app })
```

telemac:
Correct, this code does the trick, thank you :

```go
package main

import (
	"context"
	"github.com/pocketbase/pocketbase/apis"
	"log/slog"
	"time"
)

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
	defer cancel()

	app := pocketbase.New()
	app.Bootstrap()

	app.OnBeforeServe().Add(func(e *core.ServeEvent) error {
		go func() {
			<-ctx.Done()
			slog.Info("context cancelled")
			e.App.OnTerminate().Trigger(&core.TerminateEvent{App: e.App})
		}()
		return nil
	})

	_, err := apis.Serve(app, apis.ServeConfig{
		HttpAddr:        "0.0.0.0:8090",
		ShowStartBanner: true,
		AllowedOrigins:  []string{"*"},
	})
	slog.Info("apis.Serve stopped", "error", err)

}

```

PR #4999, opened 2024-05-28
Add Notion OAuth Provider

s-li1:
Hi there!

I've been enjoying Pocketbase and noticed that PRs were welcome for new OAuth providers. I thought I could contribute by adding a Notion OAuth provider 😁 

I've based most of my code from previous integrations, including [Bitbucket](https://github.com/pocketbase/pocketbase/pull/3948) and [Patreon](https://github.com/pocketbase/pocketbase/pull/3323)

I've followed the recommendations on the contributions page, including building the FE, adding tests, and running the linter.

You can create a Notion integration at: [https://www.notion.so/my-integrations](https://www.notion.so/my-integrations)

This is my first open source contribution, so let me know if there's anything wrong or missing. I'm more than happy to receive feedback! Cheers 🙌

PR #5346, opened 2024-08-12
Add monday.com as an OAuth provider

berybin:
Hey!

I've added [monday.com](https://monday.com/) as an OAuth provider.
I've intentionally kept the `.com` as per their brand guidelines [here](https://www.brand-monday.com/logo) when displaying the list of providers in the admin UI.

I've ran all tests, and lint checks.
All tests passed, and there were no additional linting issues

Info setting up a monday.com app can be found here:
- https://developer.monday.com/apps/docs/create-an-app
- https://developer.monday.com/apps/docs/oauth

I've seen [here](https://github.com/pocketbase/pocketbase/pull/4999#pullrequestreview-2081840100) that OAuth provider PRs probably won't be merged until after the refactor, so take as long as you need.
Let me know if there are any changes you'd like made before or after the refactor.

I've also verified that this works using a similar file to the one mentioned [here](https://github.com/pocketbase/pocketbase/discussions/5074):

```html
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8" />
    <title>monday.com OAuth2 test</title>
</head>
<body>
    <button id="btn">Login with monday.com</button>
    <pre id="response"></pre>

    <script type="module">
        import PocketBase from "https://cdn.jsdelivr.net/gh/pocketbase/js-sdk@master/dist/pocketbase.es.mjs"

            pb.collection('users').authWithOAuth2({
                provider: "monday",
            }).then((data) => {
            }).catch((err) => {
                console.log("Failure:", err, err.originalError);
            });
        })
    </script>
</body>
</html>
```

Let me know if you have any other feedback, cheers!


ganigeorgiev:
Thank you, I'll review and test it in more details in the next couple weeks.

_I'm not sure if it is needed but just to be safe, could you also add a check for the `enabled` user field, something like:_
```go
if !extracted.Data.Me.Enabled {
    return nil, errors.New("the Monday's user is not enabled")
}
```

ganigeorgiev:
I've tested the integration locally and it seems to work correctly.

The provider is squash-merged in the `develop` branch with some minor changes to accommodate the new internal APIs and it will be part of v0.23.0 release (_there are no ETAs yet_).

PR #5397, opened 2024-08-27
Enabling SQLite custom extensions and init hooks

rodydavis:
- Custom extensions
- Custom pragma
- Connection hooks

### DIFF
--- a/core/db_cgo.go
+++ b/core/db_cgo.go
@@ -9,9 +9,9 @@ import (
 	"github.com/pocketbase/dbx"
 )
 
-var DbCgoExtensions = []string{}
+var DBExtensions = []string{}
 
-var DbCgoPragma = `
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
I don't think this is needed. As part of the [refactoring](https://github.com/pocketbase/pocketbase/discussions/4355) and the upcoming v0.23.0 release (_there are no ETAs yet_) users will have the option to specify their own `ConnectDB` function as part of the app config. This will allow not only loading custom pragmas and extensions but also custom `database/sql` compatible drivers like SQLCipher or libSQL.

rodydavis:
That's awesome! Yeah I added this on my fork for my own use-cases and thought it would be good to share.

It was a nice exercise to learn more about the codebase 👍🏼

PR #5416, opened 2024-08-29
Fix double quoted string literal

ncruces:
Here, `'index'` is a string literal value for the `"type"` column.

In standard SQL column names are double quoted, string literals are single quoted.

Double quoted string literals are an SQLite [misfeature](https://www.sqlite.org/quirks.html#dblquote) that my driver is disables by default.

Related: https://github.com/ncruces/go-sqlite3/issues/142

### DIFF
--- a/migrations/1679943781_add_indexes_column.go
+++ b/migrations/1679943781_add_indexes_column.go
@@ -68,7 +68,7 @@ func init() {
 			TableName string `db:"tbl_name"`
 		}
 
-		indexesQuery := db.NewQuery(`SELECT * FROM sqlite_master WHERE type = "index" and sql is not null`)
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
I'll update it with the [refactoring](https://github.com/pocketbase/pocketbase/discussions/4355) but for now PocketBase currently doesn't support officially other SQLite drivers (_option to specify a custom `connectDB` method is added as part of the new app config_).

ncruces:
Well… that's disappointing.

To be clear: I wouldn't expect you to support other drivers, that's… not what this PR is about.

Using a doble quoted string literal is a bug, and this is the only place in your entire project where you made that typo. You're very consistently correct elsewhere. This is a simple, harmless, fix for that.

ganigeorgiev:
Again, this will be applied as part of the refactoring in other places where is occuring.

This migration file is not even actually used anymore and it is removed as part or refactoring.

PR #5588, opened 2024-10-02
Replace deprecated Instagram Basic Display API login with new Instagram Login method Scope, URL and Fields

pnmcosta:
Hi @ganigeorgiev 

Meta has announced that Instagram Basic Display API is being deprecated on December 4th, more info [here](https://developers.facebook.com/blog/post/2024/09/04/update-on-instagram-basic-display-api/).

The replacement is now Instagram API with Instagram Login, more info [here](https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login), or use the Facebook provider instead depending on the required use cases, more info [here](https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login/migration-guide#should-you-migrate-your-app). 

Since the provider I contributed with before was mostly used to login using the deprecated scope, and it was missing some fields that are now available in the new method I have created this PR to update it.

Please note that if someone is still using this provider for Instagram Basic Display API to login, they should still be able to, however they may need to explicitly ~~`SetScopes` of the `ProviderClient` from the `OnRecordBeforeAuthWithOAuth2Request` hook as the default has changed.~~ set the `scopes` property when initiating auth collection with the SDK's `authWithOAuth2` as per [docs](https://pocketbase.io/docs/authentication/#oauth2-integration).

Let me know if you have any questions?

### DIFF
--- a/tools/auth/instagram.go
+++ b/tools/auth/instagram.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/pocketbase/pocketbase/tools/types"
 	"golang.org/x/oauth2"
-	"golang.org/x/oauth2/instagram"
 )
 
 var _ Provider = (*Instagram)(nil)
@@ -25,16 +24,16 @@ func NewInstagramProvider() *Instagram {
 		ctx:         context.Background(),
 		displayName: "Instagram",
 		pkce:        true,
-		scopes:      []string{"user_profile"},
-		authUrl:     instagram.Endpoint.AuthURL,
-		tokenUrl:    instagram.Endpoint.TokenURL,
-		userApiUrl:  "https://graph.instagram.com/me?fields=id,username,account_type",
+		scopes:      []string{"instagram_business_basic"},
+		authUrl:     "https://www.instagram.com/oauth/authorize",
+		tokenUrl:    "https://api.instagram.com/oauth/access_token",
+		userApiUrl:  "https://graph.instagram.com/me?fields=id,username,account_type,user_id,name,profile_picture_url,followers_count,follows_count,media_count",
 	}}
 }
 
-// FetchAuthUser returns an AuthUser instance based on the Instagram's user api.
+// FetchAuthUser returns an AuthUser instance based on the Instagram's login fields.
 //
-// API reference: https://developers.facebook.com/docs/instagram-basic-display-api/reference/user#fields
+// API reference: https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login/get-started#fields
 func (p *Instagram) FetchAuthUser(token *oauth2.Token) (*AuthUser, error) {
 	data, err := p.FetchRawUserData(token)
 	if err != nil {
@@ -46,6 +45,7 @@ func (p *Instagram) FetchAuthUser(token *oauth2.Token) (*AuthUser, error) {
 		return nil, err
 	}
 
+	// @note the extracted "id" is a app scoped id, to get the actual IG ID use the RawUser's map key "user_id"
 	extracted := struct {
 		Id       string `json:"id"`
 		Username string `json:"username"`
--- a/tools/auth/instagram.go
+++ b/tools/auth/instagram.go
@@ -25,7 +25,7 @@ func NewInstagramProvider() *Instagram {
 		displayName: "Instagram",
 		pkce:        true,
 		scopes:      []string{"instagram_business_basic"},
-		authUrl:     "https://www.instagram.com/oauth/authorize",
+		authUrl:     "https://www.instagram.com/oauth/authorize?enable_fb_login=0&force_authentication=1",
 		tokenUrl:    "https://api.instagram.com/oauth/access_token",
 		userApiUrl:  "https://graph.instagram.com/me?fields=id,username,account_type,user_id,name,profile_picture_url,followers_count,follows_count,media_count",
 	}}
--- a/tools/auth/instagram.go
+++ b/tools/auth/instagram.go
@@ -45,6 +45,11 @@ func (p *Instagram) FetchAuthUser(token *oauth2.Token) (*AuthUser, error) {
 		return nil, err
 	}
 
+	// include list of granted permissions to RawUser's payload
+	if permissions := token.Extra("permissions"); permissions != nil {
+		rawUser["permissions"] = permissions
+	}
+
 	// @note the extracted "id" is a app scoped id, to get the actual IG ID use the RawUser's map key "user_id"
 	extracted := struct {
 		Id       string `json:"id"`


pnmcosta:
Additional commit includes the list of granted permissions into the raw user payload, more info [here](https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login/business-login#step-2---exchange-the-code-for-a-token). 

I figured it would be handy for anyone migrating to the new model.

Let me know if you prefer me to squash the commits?

ganigeorgiev:
Thank you! I'll have to read the linked docs in more details before merging but that will be after the upcoming v0.23.0-rc release.

From a quick glance, I'm not sure that I understand the note about the "app scoped id" and whether it would not conflict with already linked Instagram users if the old and new version returns different ids.

pnmcosta:
No worries, take your time, I'm aware this is not priority. In the meantime I'll use my fork as we're moving from Facebook login to the new Instagram method as it massively reduces onboarding steps.

> "app scoped id" and whether it would not conflict with already linked Instagram users if the old and new version returns different ids.

Honestly, I'm not sure either, I was not actually using this provider on my production app, cause we developed features since I contributed it that required the Facebook login.

One thing I know for sure "id" != "user_id" on this new method, but the majority of the Graph requests use "user_id", I think only the revoke one may require the app scoped it but I haven't gotten around to test it yet.

I'll let you know if I find out more about it and whether or not it indeed differs, or not, from the basic api login.

pnmcosta:
Ok, on https://www.instagram.com/accounts/manage_access/ with the new method and where I also see the alert:

>  The Instagram Basic Display API will be discontinued on December 4th, 2024. After that date, access to the Instagram Basic Display API and your historical usage of it will no longer be available. Download your information now if you'd like to keep it.

The User ID they include on the message specific to the account and page access is the "id" (a.k.a. app scoped id), e.g.:

> ***MYAPP*** uses your information to improve your experience. To learn more about how an app can use this information, view their Privacy Policy . If you contact this app to get support or provide feedback, they may need your User ID: ***ID***.

From this I deduct that they would be the same, I could be wrong, but I don't think they would interchange depending on whether it's the new/old method 😅 🤷

pnmcosta:
OMG! This. Is. Meta. Awkward. 😅

When authing, the "id" is 123 and the "user_id" is 999, as mentioned on my previous comment the id shown in the info message is the "id" (e.g. 123) but now that I'm testing the revoking via the manage access page the actual id i'm receiving in the request is the "user_id" (e.g. 999).

I'm going to do one other test (manual revoke from our side) and see what identifier that API request to work properly. If "user_id" I'd probably suggest disregarding "id" totally and just use "user_id", at least from the 4th of December onward.

pnmcosta:
Hi Gani,

Couple of updates on this, unfortunately the revoke feature (where we as app owners revoke access for an existing account) doesn't seem to be working with this API/Login method, I've reported the issue on Facebook's community [here](https://developers.facebook.com/community/threads/562313019465904). I'm still unsure about the whole "id"/"user_id" since I was not able to action any revoke requests successfully.

Additionally, I've updated the provider to force authentication and prevent login with Facebook as per the documentation:
image

I believe that at least for us, this is a better use of the provider. I've tried to not change these defaults and maybe use and event to customize these on per provider use, however I couldn't find one that allowed me to override the AuthUrl, the only closest was `OnRecordBeforeAuthWithOAuth2Request` but that occurs after the token exchange.

Again, please only address these when it's suitable for you, I can work around by using my fork for my needs.

Looking forward to all the other updates to PB 🙌

ganigeorgiev:
I've spent couple hours since this morning to try to understand what's going on with the new Instagram APIs and unfortunately I don't think a graceful automated transition is possible.

For the existing/old provider, based on the [API fields docs](https://developers.facebook.com/docs/instagram-basic-display-api/reference/user#fields), the id is:

> The app user's app-scoped ID (version 11.0+) or raw user ID (version 10.0 and lower). 

Since we didn't hardcoded a version in the old urls, based on the [Meta versioning docs](https://developers.facebook.com/docs/instagram-basic-display-api/overview#versions):

> If you omit the version number from your request, the request will resolve to the version specified in the App Dashboard > Settings > Advanced > Upgrade API Version section.

So we technically don't know what ids the existing linked Instagram Basic Display users actually use.
But we can make a guess that users most likely uses version 11+ APIs considering that v10.0 is pretty old (the current Meta API version is v21.0) and that the earliest currently supported API version seems to be v13.0 based on [Available Graph API versions](https://developers.facebook.com/docs/graph-api/changelog/#available-graph-api-versions).

Or in other words, I assume that all existing linked PocketBase Instagram users are all using app-scoped user ids.

And this is a problem because we can't migrate automatically existing Instagram Basic Display users to the new provider API because the OAuth2 app will be different and the new provider will return its own custom ids and PocketBase will see it as completely different account (there is no email or other stable field that we can rely on).


To make the matter worst, the new Instagram Login apis seems to allow **login only for business or creator accounts**. This means that PocketBase users who previously used the Instagram provider to login with their normal/personal Instagram account no longer will be able to do that. This is also seems to be confirmed by your [linked announcement](https://developers.facebook.com/blog/post/2024/09/04/update-on-instagram-basic-display-api/):

> After December 4th, 2024, there will no longer be a set of Instagram APIs for consumer developer apps.


**So to summarize, the new Instagram Login provider should be treated as whole new different provider and I don't know how existing users can safely transition to it.**

My suggestion is:
- keep the old implementation intact
- rename this provider to something else, for example "instagram2"
- remove the `?enable_fb_login=0&force_authentication=1` from the auth url as I'd prefer to use whatever defaults Meta apply. The auth url is public and users can adjust it manually by doing something like:
    ```js
    await pb.collection("users").authWithOAuth2({
        provider: "instagram2",
        urlCallback: (authUrl) => {
            // modify the authUrl, for example:
            const url = new URL(authUrl);
            url.searchParams.set("enable_fb_login", 0)
            url.searchParams.set("force_authentication", 1)

            window.open(url.toString(), "_blank");
        }
    })
    ```
- after December 4th, delete the old implementation and alias "instagram2" back to just "instagram"

@pnmcosta Do you have existing Instagram users and how do you plan to transition them?

pnmcosta:
Got it, I'm away this weekend and will have a look next week. Unless you pick it up. Feel free to block it onto after v0.23.

I'm not worried about my users I have a integrations collection that relates to users that references whichever oauth2 provider from pocketbase. Have a migration plan that can utilize whatever provider name.

However I was not actively using this provider, I had to move on to Facebook's and pages for features that were not available on IG basic.

ganigeorgiev:
I've merged the above suggestions in the `develop` branch. Couple minor changes:

- the provider key is `instagram2`
- we use the `user_id` as linking id instead of the app scoped id to allow keeping reference to the same Instagram user even if the developer decide to create a new Instagram OAuth2 app
- added `name` and `profile_picture_url` fields mapping

If you already use the new provider as part of your fork in production make sure to update the existing `provider` values in the `_externalAuths` table (for existing app scoped id -> user_id mappings I'm not sure what to recommend).

Feel free to test it and let me know if something need to change (I've tested with 2 different Instagram apps to verify that the same user id would be returned). The provider will part of the final v0.23.0 release (_there are no ETAs yet_).

pnmcosta:
Awesome, thank you so much Gani.

I'll be testing this in the coming weeks as we definitely want to be migrating to this new provider/API, and will report any findings.

enkyuan:
@pnmcosta @ganigeorgiev my team's looking into using the instagram api for authenticating users and were hoping for some help with the integration. we have a business account so that's not a part of the issue. the setup is a bit confusing, requiring the information highlighted by the image below:
Screenshot_2024-12-22_at_4 49 03_PM

since our development instance is hosted, the url callback is set at `https://empty-production.pockethost.io/api/oauth2-redirect`. the verify token part is confusing as a get request to the url doesn't return the expected value. if you guys don't have enough context on this, i can provide some more.

pnmcosta:
@enkyuan that's the webhooks callback configuration screen and not oauth's, they are distinct things 😅

You need to have the new Instagram product added, instructions [here](https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login), and you can the correct place for oauth config as shown below:

image

I can confirm that `/api/oauth2-redirect` is the correct route to setup on the `OAuth redirect URIs` field.

Good luck

enkyuan:
💀 my apologies for the incorrect terminology and thanks for the help lol. hopefully it works 😄

enkyuan:
just to clarify though, how are the webooks configured? it asks for a redirect url and a verification token, neither of which is explictly given. i think that's the main issue my teammate is having right now cause the oauth configuration is relatively easy compared to the webhooks setup.

pnmcosta:
You can configure [custom routes](https://pocketbase.io/docs/go-routing/#registering-new-routes) with PB for a `GET /webhook` verification handler that returns a specific payload (used only when verifying a callback on FB app) and a  `POST /callback` handler that validated and processes payloads, more detailed info on [here](https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login/webhooks#create-endpoint).

[This](https://github.com/pnmcosta/go-meta-webhooks) could probably help if you want to validate, parse and handle webhook payloads, but I don't support all types yet, documentation may not be up to date however I've been running it on my app for the past year or so without major issues.

enkyuan:
appreciate the response. i'll lyk if it works out :)

PR #5683, opened 2024-10-15
Fix EmailTemplate Documentation

zerbzman:
Updated return documentation to follow what is actually returned by the function.

I'm not 100% sure my typescript fix is correct. I ran into this today in the docs and thought I would at least bring it to your attention. Thanks

### DIFF
--- a/plugins/jsvm/internal/types/generated/types.d.ts
+++ b/plugins/jsvm/internal/types/generated/types.d.ts
@@ -18678,9 +18678,10 @@ namespace settings {
  interface EmailTemplate {
   /**
    * Resolve replaces the placeholder parameters in the current email
-   * template and returns its components as ready-to-use strings.
+   * template and returns its components as ready-to-use strings in an array. The first element 
+   * being the subject, second the body, and third the action URL.
    */
-  resolve(appName: string, appUrl: string, token: string): string
+  resolve(appName: string, appUrl: string, token: string): Array<string>
  }
 }
 


ganigeorgiev:
The documentation is autogenerated from the Go source.

This seems to be a bug with [tygoja](https://github.com/pocketbase/tygoja) as it probably fails to properly parse the combined `(subject, body, actionUrl string)` return declaration.

I've added it in my todo and will consider checking sometime later this week.

ganigeorgiev:
I've pushed a fix for the [tygoja bug](https://github.com/pocketbase/tygoja/commit/d6ff411a0f7525444cbef4c8c4f847b0a422b65c#diff-845a10fa4d109ac9211cb53a8d1540b850701f3d97f330ced94903b8ad23559e).

The PocketBase JSVM types will be updated with the next major release soon.

_As a side-note, with the upcoming v0.23.0 release the global app settings email templates were removed and now each auth collection have its own mail templates configuration (the structure is similar to the old but `actionURL` is no longer available as a field and instead its value is migrated to be embedded directly into the `body` text)._

kendallroth:
> As a side-note, with the upcoming v0.23.0 release the global app settings email templates were removed and now each auth collection have its own mail templates configuration (the structure is similar to the old but actionURL is no longer available as a field and instead its value is migrated to be embedded directly into the body text).

Apologies if I missed this somewhere in the release notes, but I was trying to track down when this changed (as during upgrade to v23 I noticed the `actionUrl` fields were no longer enabled (had been using them previously). I've noticed a couple things (this, as well as token settings moving from global to collections) that I can't find documented in `CHANGELOG`, release notes, or migration guide, which made me wonder if I maybe was just overlooking something else?

ganigeorgiev:
@kendallroth The settings change is not documented properly (it is just briefly mentioned in the [v0.23.0 CHANGELOG](https://github.com/pocketbase/pocketbase/blob/master/CHANGELOG.md) under the Web APIs section).

Specifically for the email template settings they are all under their related auth collection. For a list with the collection fields you can check https://pocketbase.io/docs/go-collections/#collection-properties. All `*Template` suffixed fields share the same `{subject, body}` struct, meaning that `actionUrl` is no longer available and the automated v0.23 db migration would attempt to replace its `{ACTION_URL}` placeholder with its direct string equivalent in the template, for example `{APP_URL}/_/#/auth/confirm-verification/{TOKEN}`.

If you are still not able to resolve your issue, please open a Q&A discussion with the code you are having difficulties migrating to v0.23.

kendallroth:
Gotcha, thanks 👍. I had found the "moved" email template settings, but had noticed that `actionUrls` were no longer available (had also come across the Go migration that handled updating it). I'm not entirely certain _why_ I was relying on the `ACTION_URL` instead of simply modifying the URL in the template (and using `APP_URL` placeholder), which will be quite easy to move to. Thanks for the explanation of why this has occurred!

image

PR #5829, opened 2024-11-09
add wakatime-oauth

tigawanna:
Hello,

I have developed an integration for the Wakatime API's OAuth provider for PocketBase, as I could not find any existing helpers for this purpose. My implementation is largely based on the Notion integration.

I have followed all the recommendations outlined on the contributions page, including building the frontend, adding tests, and running the linter. My implementation did not flag any new issues.

I have also tested the integration on a local build, and it appears to be functioning correctly. I would appreciate any feedback you may have.

Thank you.


tools/auth/wakatime.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if err := json.Unmarshal(data, &extracted); err != nil {
+		return nil, err
+	}
+
+	user := &AuthUser{
+		Id:           extracted.Data.Id,
+		Name:         extracted.Data.DisplayName,
+		Username:     extracted.Data.Username,
+		Email:        extracted.Data.Email,
```
ganigeorgiev, line 70: Assign the email field only after ensuring that `is_email_confirmed` is true (_the field should exist at least based on example response from https://wakatime.com/developers#users_).

tools/auth/wakatime.go
```diff
@@ -0,0 +1,83 @@
...
+	type Data struct {
+		Id                      string    `json:"id"`
+		DisplayName             string    `json:"display_name"`
+		Username                string   `json:"username"`
+		Email                   string    `json:"email"`
+		Photo                   string    `json:"photo"`
+	}
+	extracted := struct {
+		Data Data `json:"data"`
+	}{}
```
ganigeorgiev, line 60: Since this struct is not used anywhere else it can be "inlined", aka.:

```go
extracted := struct {
    Data struct {
    ...
    } `json:"data"`
}{}
```
(_also make sure that your editor runs gofmt because there seems to be extra spaces between the field name and type_)

tools/auth/wakatime.go
```diff
@@ -0,0 +1,83 @@
...
+	BaseProvider
+}
+
+// NewWakatimeProvider creates a new Wakatime provider instance with some defaults.
+func NewWakatimeProvider() *Wakatime {
+	return &Wakatime{BaseProvider{
+		ctx:         context.Background(),
+		displayName: "Wakatime",
+		pkce:        false,
+		scopes:      []string{"read_summaries","email"},
```
ganigeorgiev, line 31: Let's leave the minimal scopes for now, aka. just the `email` because the provider also seems to suggest specifying only the summaries that are needed:

> Consider instead requesting scopes for only the summaries you need. For ex: scope=read_summaries.languages,read_summaries.editors to only request access to the user’s language and editor summaries and stats.

Users can specify these and any other scopes from the client-side with the `authWithOAuth2` call.

tools/auth/wakatime.go
```diff
@@ -0,0 +1,83 @@
...
+		authURL:     "https://wakatime.com/oauth/authorize",
+		tokenURL:    "https://wakatime.com/oauth/token",
+		userInfoURL: "https://wakatime.com/api/v1/users/current",
+	}}
+}
+
+// FetchAuthUser returns an AuthUser instance based on the Wakatime's user API.
+//
+// API reference: https://wakatime.com/developers#authentication
```
ganigeorgiev, line 40: Replace the url with the one pointing directly to the user apis, aka. `https://wakatime.com/developers#users`.

ui/public/images/oauth2/wakatime.svg
ganigeorgiev, line 1: The SVG in the GitHub preview seems to be grey/white but I'm not sure if that will be visible in the PocketBase UI. Please replace it with one of the [darker variants](https://wakatime.com/legal/logos-and-trademark-usage) (_there is no official support for dark mode at the moment_).

tools/auth/wakatime.go
```diff
@@ -0,0 +1,83 @@
...
+type Wakatime struct {
+	BaseProvider
+}
+
+// NewWakatimeProvider creates a new Wakatime provider instance with some defaults.
+func NewWakatimeProvider() *Wakatime {
+	return &Wakatime{BaseProvider{
+		ctx:         context.Background(),
+		displayName: "Wakatime",
+		pkce:        false,
```
ganigeorgiev, line 30: Try to change the default `pkce` option to `true`, unless Wakatime explicitly requires to be disabled _(so far only LinkedIn OIDC has been reported to have this quirk which is also the reason why the field was added in the first place)_.

In general most providers usually will just ignore the extra PKCE query parameters if they don't support it.

ganigeorgiev (changes requested):
Thank you! I'll review and test the integration in more details sometime next week but for now I left some nitpicks. Feel free to check them when you find time.

tigawanna:
added the requested changes

ganigeorgiev:
I've tested the provider and squash merged the changes locally with some minor adjustments. It will be available with the next v0.23.0 release (_there are no ETAs yet_).

PR #5856, opened 2024-11-13
New template cache request option

chnfyi:
Hi,

This PR aims to improve developer experience when working with templates rendering.

JSVM restart on changes and force-cache-only loading make development slow: see #3969.

I imagine as well that there could be some production mode edge cases where some template cache request options could be useful.

So i propose :
- A new cache request option on Load* methods to disable cache use per request
- A new ForceCache method to disable default registry systematic cache use
- A new ClearCache method to reset registry cache

However I am not sure if unwatching _views_ directory to deactivate JSVM reload while in dev mode is an appropriate solution. I made the modification because i think it's better than moving _views_ folder somewhere other than in `__hooks`.

**Implementation details:**

```go
registry := template.NewRegistry()

// Now first parameter is a []string
// Second parameter (cache use) is a bool and is optional
// Load* methods still caches (if not already) by default
html1, err := registry.LoadFiles([]string{
	"layout.html",
	"content.html",
} //, true
).Render(map[string]any{"name": "John"})


// Turn off cache use per request
// Files will be newly parsed and cache will be updated
html2, err := registry.LoadFiles([]string{
	"layout.html",
	"content.html",
}, false).Render(map[string]any{"name": "Jane"})


// Turn off registry default systematic reuse of already parsed and cached files
registry.ForceCache(false)


// With ForceCache(false) previously called, now you can invoke without bool option
html3, err := registry.LoadFiles([]string{
	"layout.html",
	"content.html",
}).Render(map[string]any{"name": "Juan"})


// New cache clearing method
registry.ClearCache()

``` 

With JSVM :
```javascript
const html = $template.loadFiles([
	"views/layout.html",
	"views/content.html",
]).render({"name": "John"})
``` 

`go test ./tools/template` is ok

### DIFF
--- a/plugins/jsvm/internal/types/generated/types.d.ts
+++ b/plugins/jsvm/internal/types/generated/types.d.ts
@@ -3332,17 +3332,17 @@ namespace filesystem {
  * ```
  * 	registry := template.NewRegistry()
  * 
- * 	html1, err := registry.LoadFiles([
+ * 	html1, err := registry.LoadFiles([]string{
  * 		// the files set wil be parsed only once and then cached
  * 		"layout.html",
  * 		"content.html",
- * 	]).Render(map[string]any{"name": "John"})
+ * 	}).Render(map[string]any{"name": "John"})
  * 
- * 	html2, err := registry.LoadFiles([
+ * 	html2, err := registry.LoadFiles([]string{
  * 		// reuse the already parsed and cached files set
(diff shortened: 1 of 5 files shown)


ganigeorgiev:
No sorry, there are no plans to change the current template helpers.

I'm also against adding knobs and special hardcoded directories when I personally don't or understand the use case.

PR #5885, opened 2024-11-17
Ensure all directories are included when using DeletePrefix

nehmeroumani:
Fixes an issue in the DeletePrefix logic where directories containing only subdirectories (no files) might be skipped.

### DIFF
--- a/tools/filesystem/filesystem.go
+++ b/tools/filesystem/filesystem.go
@@ -322,6 +322,32 @@ func (s *System) DeletePrefix(prefix string) []error {
 	// try to delete the empty remaining dir objects
 	// (this operation usually is optional and there is no need to strictly check the result)
 	// ---
+	
+	// If a directory contains only subdirectories and no other files, it might be skipped in certain cases.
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Could you elaborate why this is needed?

A test case illustrating the problem would also help.

ganigeorgiev:
No, sorry, I don't think this is needed. It is too errors prone and it seems to rely that `Delete` will throw an error in case a directory is not empty.

I'm not sure in what situation the problem you are experiencing is happening but the directory delete operation is considered optional and I don't want to complete unnecessary unless it is causing an actual issue.

Feel free to elaborate why this is needed backed by an automated test but for now this remains out of the scope.

nehmeroumani:
This issue occurs when using the Filesystem instance in custom code. For example, attempting to use DeletePrefix to remove a directory containing only subdirectories with no other files can trigger this behavior. Considering that this is likely an uncommon scenario and empty directories may not be an issue for all users, I agree that addressing it adds unnecessary complexity. Keep up the great work!

PR #5909, opened 2024-11-21
Add Linear OAuth Provider

chnfyi:
Hi,

I've added Linear as an OAuth provider, based on previous OAuth integrations.

Testing locally, I've checked everything several times: PR code, PocketBase UI, Client integration, Linear dashboard (client ID, client secret). But i'm still getting a _"Failed to fetch OAuth2 token"_ error on `authWithOAuth2` with `oauth2: "invalid_grant" "Invalid grant: code verifier is invalid"` as details.

**Note**: 
- I've set `http://127.0.0.1:8090/api/oauth2-redirect` callback URL into my Linear App. 
- I can authenticate into Linear and accept integration through the popup window then see the URL to be `oauth2-redirect-success` just before being automatically closed.

If it doesn't work for you also, let me know how i can fix it.


ganigeorgiev:
If the integration doesn't work then it is not ready for a PR.

I'm finalizing the v0.23.0 release and don't have time at the moment to read the Linear documentation and find out why it fails.

I recommend to test it with a plain `pb_public/index.html` file like:
```html
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8" />
    <title>OAuth2 test</title>
</head>
<body>
    <button id="btn">Login</button>
    <pre id="response"></pre>

    <script type="module">
        import PocketBase from "https://cdn.jsdelivr.net/gh/pocketbase/js-sdk@develop/dist/pocketbase.es.mjs"

        pb.authStore.clear();

            pb.collection("users").authWithOAuth2({
                provider: "linear",
            }).then((data) => {
            }).catch((err) => {
            });
        })
    </script>
</body>
</html>
```

If it still doesn't work, you can experiment switching the `pkce` option to `false`.

chnfyi:
Now it works !

Switched `pkce` option to `false` and corrected the way to retrieve user info to fit Linear GraphQL API.

ganigeorgiev:
Earlier today I contacted the Linear support to ask for the PKCE and they confirmed that it is not supported at the moment.

I've rebased and squash merged the changes locally with some minor adjustments:
- added check for the `active` user field to prevent authentication for disabled (suspended) accounts
- swapped `name` and `displayName` because the latter is the nickname/username that is unique per organization

The provider will be available with the final v0.23.0 release soon.

PR #5916, opened 2024-11-22
Fix (UI): do not show batch tab on view collection

skowrons:
## Description

Batch tab would be shown in the docs panel for views.

## Bug

Bildschirmfoto 2024-11-22 um 14 55 47

## After

Bildschirmfoto 2024-11-22 um 14 55 38


ganigeorgiev:
Thank you. I'll update manually the API Preview docs with the final v0.23 release as the ui/dist also need to be regenerated.

PR #6028, opened 2024-12-04
add geolocation field type `core.FieldPoint` and internal `types.Point`

### LINKED ISSUE #21: Add gps/location field type
Add new field type allowing to set GPS coordinates (or location address).

Leaflet map with drag&drop pin(s) could be used in the admin UI.

Ideally the field should have a configuration for supporting single or multiple values _(and eventually support for a simple distance/proximity queries)_.

cowtoolz:
`core.FieldPoint` is a field type "Point" that represents a decimal-degree latitude & longitude, serialized as a comma separated string containing a pair of float64s (i.e. `-23.12414, 124.52131`) that are stored in a new internal type, `types.Point` which stores the corresponding float64s. Using float64s gives us incredibly low complexity with a worst-case precision of approximately 3.16 nanometers (on Earth).

Geolocation queries are not yet implemented, but could be handled easily by the consumer. There would be significant value in adding simple point-based queries, though, and some discussion on how this is implemented would be necessary. Checking whether a point lies in/out of an array of points (i.e. a geofence) or in/out of a radius around a point would provide enough for the vast majority of use cases; both of these could be implemented in a handful of ways, though, so some input would be nice.

Point fields are not currently implemented in the UI. I haven't looked at this yet, but frontend is not my wheelhouse so I will likely need some help here.

Fixes #21

### DIFF
--- a/core/field_point.go
+++ b/core/field_point.go
@@ -1,0 +1,135 @@
+package core
+
+import (
+	"context"
+	"fmt"
+
+	validation "github.com/go-ozzo/ozzo-validation/v4"
+	"github.com/pocketbase/pocketbase/core/validators"
+	"github.com/pocketbase/pocketbase/tools/types"
+)
+
+func init() {
+	Fields[FieldTypePoint] = func() Field {
+		return &PointField{}
+	}
+}
+
+const FieldTypePoint = "point"
+
+var _ Field = (*PointField)(nil)
+
+// PointField defines "point" type field to store a single [types.Point] value.
+//
+// The respective zero record field value is the zero [types.Point].
+type PointField struct {
+	// Name (required) is the unique name of the field.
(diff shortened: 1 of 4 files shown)


weebney:
Also, I just chose the name "Point" because it was short and made enough sense to me. I'm sure a better name is out there—I'm not married to it.

ganigeorgiev:
Thank you for the effort but the implementation is incomplete and as mentioned in the linked issue, this for now remains a very low priority. You are free to create a PoC but I don't have the time to invest to review it.

Please make sure to check the [contribution note in the README](https://github.com/pocketbase/pocketbase?tab=readme-ov-file#contributing)

weebney:
Yeah, I was just leaving this as a draft for now if anyone else wanted to contribute.

PR #6041, opened 2024-12-06
Add Webex OAuth2 provider

kaechele:
This adds Webex as an OAuth provider. Should be pretty straightforward.

Let me know if you have any questions.


tools/auth/webex.go
```diff
@@ -0,0 +1,80 @@
...
+		Username:     extracted.UserName,
+		AvatarURL:    extracted.Avatar,
+		RawUser:      rawUser,
+		AccessToken:  token.AccessToken,
+		RefreshToken: token.RefreshToken,
+	}
+
+	if len(extracted.Emails) == 1 {
+		user.Email = extracted.Emails[0]
+	}
```
ganigeorgiev, line 75: If users can have more than one email, please change the if  condition to `len(extracted.Emails) > 0`.

Also could you confirm that the returned emails are verified?
If they are not verified then we need to skip the assignment because the email is not safe for linking.

tools/auth/webex.go
```diff
@@ -0,0 +1,80 @@
...
+
+	rawUser := map[string]any{}
+	if err := json.Unmarshal(data, &rawUser); err != nil {
+		return nil, err
+	}
+
+	extracted := struct {
+		Id          string   `json:"id"`
+		DisplayName string   `json:"displayName"`
+		UserName    string   `json:"userName"`
```
ganigeorgiev, line 55: There doesn't seem to be a `userName` field listed in the Webex `/v1/people/me` docs. Maybe it should be `nickName`?

ganigeorgiev (changes requested):
Thank you! I'll review and test the integration sometime later next week, but I left some nitpicks that you could check when you find time.

tools/auth/webex.go
kaechele: Yes, users can have more than one email, but the simple `/v1/people` API only uses one email address. I would agree that the documentation isn't really clear on this for the `GET` verb, but for the `POST` and `PUT` verbs it states:
> `emails`
> array[string] **required**
>
> The email addresses of the person. Only one email address is allowed per person.

We do have an SCIM 2.0 compliant API at [`/v1/scim2-user/get-me`](https://developer.webex.com/docs/api/v1/scim2-user/get-me) that would have a more definitive representation of multiple emails, so it might be worth using that instead.
However, a Webex user can only have multiple email addresses if they use an external authentication source. If they signed up independently through webex.com they can only have one email address.

I can confirm that the email is verified.

- Direct sign-ups through webex.com do not get to the stage where they can create a password and complete account setup without first verifying their email address.
- For OAuth authenticated users must use the same email address their OAuth provider returns for their email. So if you sign up to Webex using Google OAuth you will have to use your `@gmail.com` address for you Webex username (or the email you use as your Google account email, if your account doesn't have Gmail/is a Google Workspace account with custom domain, at which point you will have to have verified it with Google previously).
- For enterprise users the email is whatever their authentication source provides. So I would say it is safe to assume it is verified on their end as well.

It is safe to say that if the user ever arrives at the point where they can trigger PocketBase to make this API call there will be a verified email that is returned.

tools/auth/webex.go
kaechele: That seems to be an omission in the documentation then. I will let the product team know.
The `userName` field definitely exists for both internal auth (direct sign-ups) and external auth users in the response and is generally the user's email address.

Kvm4 (approved):

Davidak643h (approved):

ganigeorgiev:
Sorry for the delay.

I started reading the Webex docs in more details and I don't think a custom OAuth2 provider is needed in our case because Webex seems OpenID Connect (OIDC) compatible.

Or in other words, users that want to authenticate with Webex OAuth2 can simply use the generic PocketBase OIDC provider. The configuration endpoint are similar to the suggested implementation, except the userinfo, which should be https://webexapis.com/v1/userinfo (the discovery endpoint is available in https://webexapis.com/v1/.well-known/openid-configuration).

I've tested the OIDC provider with Webex earlier today and it works correctly for me with the following `pb_public/index.html`:

```html
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8" />
    <title>OAuth2 test</title>
</head>
<body>
    <button id="btn">Login</button>
    <pre id="response"></pre>

    <script type="module">
        import PocketBase from "https://cdn.jsdelivr.net/gh/pocketbase/js-sdk@v0.22.1/dist/pocketbase.es.mjs"

        // make sure to add "http://127.0.0.1:8090/api/oauth2-redirect" as redirect url in the Webex app settings

        pb.authStore.clear();

            pb.collection("users").authWithOAuth2({
                provider: "oidc",
            }).then((data) => {
            }).catch((err) => {
            });
        })
    </script>
</body>
</html>
```

kaechele:
No worries!

I really value the time you put into this. Thank you for that. I appreciate that you took the time to provide a usage example as well.

And sorry for not figuring this out myself earlier. But it's great news that we can just use the OIDC provider.
It seems that our `/userinfo` endpoint does not expose a value that can be used for `avatarUrl`. But I guess this can be worked around by calling the `/identity/scim/v2/Users/me` endpoint separately and setting the `avatarUrl` from that.

PR #6045, opened 2024-12-07
Replace math/rand with math/rand/v2

apocelipes:
"math/rand/v2" is much faster than "math/rand" and has similar APIs. 
The go.mod file has specified that the minimum version of go is v1.23, so we can safely replace all "math/rand" with "math/rand/v2".

### DIFF
--- a/tools/security/random.go
+++ b/tools/security/random.go
@@ -3,7 +3,7 @@ package security
 import (
 	cryptoRand "crypto/rand"
 	"math/big"
-	mathRand "math/rand" // @todo replace with rand/v2?
+	mathRand "math/rand/v2"
 )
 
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Thank you but I still haven't got the time to explore the v2 API changes and I don't want to merge it right now.

apocelipes:
All right. But the only big change is the random source APIs, other things only changed their names...

ganigeorgiev:
That's fine but I don't need help with this or any of the tasks in the roadmap.

I appreciate the effort but please read the [contribution note in the README](https://github.com/pocketbase/pocketbase?tab=readme-ov-file#contributing).

apocelipes:
Thanks for replying, I think it is wise to keep the working code as-is. Sorry for the code churn.

PR #6063, opened 2024-12-08
fix XFF header typo

btwiuse:
"X-Forward-For" should be `X-Forwarded-For`

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For


btwiuse:
There are other two cases of the typo left unmodified, one in `CHANGELOG_16_22.md`, another in `ui/dist/assets/index-DKhIJ63w.js`

image

Please help me decide if those should also be corrected.

ganigeorgiev:
Thank you. I'll apply the change manually as the generated `.d.ts` shouldn't be manually updated and the `ui/dist` need to be regenerated.

Regarding the v0.22 changelog - it should remain as it is because it represents the correct state of the respected tagged version. With that said, I'll backport the typo fix in the v0.22 maintenance branch but will not rush another release for now as it is not critical or a blocker.

PR #6144, opened 2024-12-18
Fix comment typo at event_request.go

ahmedabdou14:
`my another handler` -> `by another handler`


ganigeorgiev:
Thank you. I've applied it manually in the `develop` branch together with some other minor code comment changes since the jsvm types also had to be regenerated.

PR #6155, opened 2024-12-20
Exclude health checks from activity log

gabe565:
Hey! I'm loving v0.23, but I've noticed my uptime monitors are causing a lot of logs for the `/api/health` endpoint. This PR sets `requestEventKeySkipSuccessActivityLog` to exclude these.


ganigeorgiev:
The health endpoint is public and it is better to have logs by defalt so that you know who invoked the action.

You can:
- change the min log level to something > 0
- add a middleware with Go or JS that will exclude it; I haven't tested them but here is an example:
    JS:
    ```js
    // pb_hooks/main.pb.js
    routerUse((e) => {
        if (e.request.url.path == "/api/health") {
            return $apis.skipSuccessActivityLog().func(e);
        }

        return e.next()
    })
    ```
    
    Go:
    ```js
    e.Router.BindFunc(func(e *core.RequestEvent) error {
        if e.Request.URL.Path == "/api/health" {
            return apis.SkipSuccessActivityLog().Func(e)
        }
        return e.Next()
    })
    ```

As side-note, keep in mind that the `develop` branch is not safe to be used in production environment and the implemented features that are not in `master` can change without notice, so I'd recommend to avoid using it for now and switch to the latest stable v0.23.12.

gabe565:
@ganigeorgiev Ha, you closed this as I was rebasing to merge into the `develop` branch, ignore the random commits. I know it's a public API, but I just can't see those logs being useful. I'll add the middleware to all of my PocketBase apps. Thank you!

gabe565:
@ganigeorgiev The Go middleware works fine!

PR #6275, opened 2025-01-10
DOC - Update list of API SDK clients

Stoufiler:
In order to reflect what is said in public documentation we should have the same values as exposed on this page : https://pocketbase.io/docs/client-side-sdks/


ganigeorgiev:
Thank you. The  `/client-side-sdks` page will change in the near future and I'll consider syncing the readme with it but for now lets keep it as it is.

PR #6281, opened 2025-01-11
Update db.go

svobol13:
Fixed a bug where multiple fields starts same.

I got two fields: price and price_id. If there is a problem with price_id field, then also price field matches and program incorrectly reports error at both fields. Adding spaces fixed my problem.

### DIFF
--- a/core/validators/db.go
+++ b/core/validators/db.go
@@ -64,7 +64,7 @@ func NormalizeUniqueIndexError(err error, tableOrAlias string, fieldNames []stri
 
 		for _, name := range fieldNames {
 			// blank space to unify multi-columns lookup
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Thank you. I've added it in my todo to check it locally a little bit later and will add a test for this but for now lets leave it as it is.

svobol13:
Thanks. Steps to reproduce:
1. Create collection with two fields: price and price_id
2. Add unique constraint to price_id
3. Insert programaticaly 2 records with same price_id.
4. Second record fail with two validation errors that says that both fields price and price_id violates unique constraint. Expected is only one error.

PR #6284, opened 2025-01-12
Fix typo on marshal(l)ed

yall:
Fixed a typo on *marshalled* that is spelled *marshaled* everywhere.

Nothing much 😅


ganigeorgiev:
I've added it in the roadmap to apply it manually with the next v0.25 release since it would be a breaking change and while maybe inconvenient I don't think it is critical (_and eventually maybe keep the public methods with the wrong spelling for couple more versions and soft-deprecate them with calls to the correct ones_).

ganigeorgiev:
Actually after searching for the correct spelling I'm not sure if this needs to change as it seems that the single "l" is considered also valid (at least based on the [WikiPedia article](https://en.wikipedia.org/wiki/Marshalling_%28computer_science%29)) (there are also several Google entries for "marshaled" vs "marshalled" and both seems to be valid).

I'm not a native speaker, so I'm not really sure what is the correct spelling but if spelled with a single "l" is also valid I'll just keep it as is then and will ensure that everywhere in the code is consistent (_the double "ll" version is used only in a single place as a comment_).

yall:
I'm not a native speaker neither, but it looked weird because I was used to the double L spelling.
You're absolutely right! Also Golang documentation uses "marshaled" spelling: [https://pkg.go.dev/encoding/json](https://pkg.go.dev/encoding/json)
Sorry for the useless PR 😅

PR #6312, opened 2025-01-19
make compatible changes for exist checking with sql driver "rqlite"

### LINKED ISSUE #6304: check Exist of row should NOT use bool to scan
if not use default sqlite driver to run this query or lots of check like here https://github.com/pocketbase/pocketbase/blob/231769501177e5e76b37bb2d0ace0ef35c8096dc/core/migrations_runner.go#L270

```
                        var exists bool

			err := db.Select("count(*)").
				From("_migrations").
				Where(dbx.HashExp{"file": "1640988000_aux_init.go"}).
				Limit(1).
				Row(&exists)
```

will raise a panic:

`panic: sql: Scan error on column index 0, name "count(*)": sql/driver: couldn't convert 1 (float64) into type bool`

this will pass:
```
var exists int

....query

exists > 0

```

trawor:
fix https://github.com/pocketbase/pocketbase/issues/6304

### DIFF
--- a/apis/realtime.go
+++ b/apis/realtime.go
@@ -722,7 +722,7 @@ func realtimeCanAccessRecord(
 		return false
 	}
 
-	var exists bool
+	var exists int
 
 	q := app.DB().Select("(1)").
 		From(record.Collection().Name).
@@ -739,5 +739,5 @@ func realtimeCanAccessRecord(
 
 	err = q.Limit(1).Row(&exists)
 
-	return err == nil && exists
+	return err == nil && exists > 0
 }
--- a/apis/record_crud.go
+++ b/apis/record_crud.go
@@ -314,9 +314,9 @@ func recordCreate(optFinalizer func(data any) error) func(e *core.RequestEvent) 
 
 					resolver.UpdateQuery(ruleQuery)
 
-					var exists bool
+					var exists int
 					err = ruleQuery.Limit(1).Row(&exists)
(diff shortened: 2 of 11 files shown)


ganigeorgiev:
@trawor Could you provide a code sample how do you load rqlite?

ganigeorgiev:
I've added task in the roadmap to consider replacing the bool scans with int for broader compatibility with the next v0.25.0 release as there are already minor changes to some of the targeted queries in the `develop` branch.

With that said, assuming that you want to use `rqlite/gorqlite/stdlib`, considering the [outlined driver implementation caveats](https://github.com/rqlite/gorqlite?tab=readme-ov-file#driver-for-databasesql) I'm not sure even with that change whether everything will work as expected (for example the batch APIs failure handling).

_There is no information about your use case and how do you plan using rqlite with PocketBase (_PocketBase can't scale on more than 1 server and the database is not the only limitation_),
so my recommendation from the docs stands - use a custom driver if you really think that you'll benefit from its features, otherwise if you are note sure - I recommend sticking with the default one._

trawor:
Thanks for explaining, 
Yes, I'm making a progress with rqlite features to scale up (imagine several offices in different locations, and try to keep LAN accessing service available even without internet, because the service has the control of WiFi auth )

And your consideration of testing makes scenes, I will try to test on my own dev env, till we can start 0.25

Issue #6433, opened 2025-02-09
Bug: Realtime Delete Broadcast Silently Fails for Custom Record Structs


Snonky:
## Description

Hi! During my development of https://github.com/pocketbase/pocketbase/issues/6388 I encountered a bug associated with custom record model structs.

The realtime messages for delete actions are not sent when `app.Delete` is called with a custom `core.RecordProxy` struct. The fail is silent, it causes no error.

## System

- Pocketbase v0.25.0
- windows
- amd64

## Analysis

I looked into it and believe I found the critical code snippets:

### `OnModelAfterDeleteSuccess` handler

https://github.com/pocketbase/pocketbase/blob/d3aac570b2922f22b8d09f8903a529d585b48cff/apis/realtime.go#L375-L394

Here the broadcast of the delete event happens after `realtimeResolveRecord` provides a `*core.Record` on line 378. This is where nil is returned in the buggy case. 

### `realtimeResolveRecord` with a custom record model

https://github.com/pocketbase/pocketbase/blob/d3aac570b2922f22b8d09f8903a529d585b48cff/apis/realtime.go#L418-L445

Here on line 440 the function attempts to fetch the record from the `app` to resolve it. This fails because the delete is already completed in `OnModelAfterDeleteSuccess` handlers.

## Possible fix

Maybe it can be fixed by replacing line 421 of `realtimeResolveRecord` with a type switch that checks for `core.RecordProxy` implementors and then using the interface's `ProxyRecord` function to resolve the realtime record.

ganigeorgiev:
I've submitted a "fix" in `master` and it will be available with the next minor release sometime later next week.

_We currently don't have tests for the realtime messaging as it is a little troublesome to setup but I'll think a little more on it next week to see if we can create some helpers for that or at least manually verify that it works correctly for all scenarios._

ganigeorgiev:
Everywhere where `RecordProxy` is resolved is expected to be properly loaded (similar to calling `core.NewRecord()` with a valid Collection model).

I'll consider with the next release to add guards specifically for this but there are many places where this may not be easily done or will further complicate the code and if it is getting misued or starting to cause too much issue I'll just ended up removing the `RecordProxy` abstraction altogether.

PR #6434, opened 2025-02-09
Bug Fix for Realtime Delete Events Failing when Using Custom Record Model Structs

Snonky:
PR for https://github.com/pocketbase/pocketbase/issues/6433

Sorry for not including a test, I do not know how to test for this


ganigeorgiev:
Thank you but it is not enough as it will solve the issue only partially. I'll try to submit a more general fix a little bit later.

PR #6456, opened 2025-02-13
added dockerfile

huuthangntk:


ganigeorgiev:
I don't think this is needed.

Please don't spam and read the contribution note in the README!

PR #6657, opened 2025-03-28
Support email_verified bool as string for AWS Cognito OIDC

dwendt:
Hello! AWS Cognito is an OIDC provider that has a single quirk that makes it unique, and sometimes seen across other OIDC providers that use AWS under the hood to implement their OIDC. It returns the "email_verified" portion of userinfo as a string such as `"true"` or `"false"`. 

Most libraries support a string-to-bool for this field as a result of this common spec violation https://github.com/zitadel/oidc/issues/137

This PR is a fix, that I confirmed works for AWS Cognito myself, but open to change to however you'd prefer to handle it. I recognize we might permit other fields to be invalid in this approach, but wanted to keep the change minimal.

### DIFF
--- a/tools/auth/oidc.go
+++ b/tools/auth/oidc.go
@@ -94,8 +94,17 @@ func (p *OIDC) FetchAuthUser(token *oauth2.Token) (*AuthUser, error) {
 		Email         string `json:"email"`
 		EmailVerified bool   `json:"email_verified"`
 	}{}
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
I've pushed a slightly different implementation in `master` and it will be available with the next minor release sometime later this weekend (_will consider backporting it also in the `v0.22_maintanence` branch_).

PR #6742, opened 2025-04-19
Add WeChat OAuth2 provider

fondoger:
In China, more than 1 billion people use WeChat. And **Login with WeChat** is used much more frequently than **Login with Apple** for Chinese mobile users.

This PR adds WeChat OAuth2 provider to pocketbase.

Note that for WeChat, **mobile** and **web** login flow uses two different scopes. So I split WeChat OAuth2 provider into `wechat_mobile` and `wechat_web`.
image


ganigeorgiev:
Thank you for the effort but I don't want to merge it for now as it would be difficult for me to maintain it. See also https://github.com/pocketbase/pocketbase/pull/2865#issuecomment-1664480126.

Note that while the Dashboard UI at the moment is not extendable, users can register external providers programmatically by modifying the exported `auth.Providers` and manually add the provider to the collection model options.

ganigeorgiev:
@fondoger No sorry, the reasons are explained in the previous PR and I don't want to merge it for now.

Again, if it is critical for your use case you can register the provider programmatically with Go.

Issue #825, opened 2022-10-23
Files: Thumb not working with WEBP format


Siumauricio:
Hi, I was testing to upload and resize WEBP images with the thumb property, the file upload is working fine, but the resize is not working I was checking the code probably this is the issue, it's just taking these formats and not WEBP.

https://github.com/pocketbase/pocketbase/blob/efda3d5a0b7e2dfd6ced4e1217cccfb9497a9759/apis/file.go#L12

https://github.com/pocketbase/pocketbase/blob/efda3d5a0b7e2dfd6ced4e1217cccfb9497a9759/apis/file.go#L71

ganigeorgiev:
Unfortunately It's not just the file extension that is missing.

Webp is not supported at the moment because the image library that we use, [disintegration/imaging](https://github.com/disintegration/imaging), doesn't support webp (mostly due to the lack of native go webp encoding - https://github.com/golang/go/issues/45121; most webp go packages seems to rely on `libwebp` which requires cgo and I want to avoid that).

There are workarounds, like decoding the uploaded image and converting it to jpg/png thumb, but for now this remains out of the scope for v1.0.

KnockOutEZ:
@ganigeorgiev can I look into this and try to solve this without cgo?

ganigeorgiev:
@KnockOutEZ I'm open for suggestions but please first outline some of your planned implementation details to avoid investing unnecessary time and efforts in something that may not get merged.

Wikidepia:
I stumbled upon this issue, just want to add that there are CGo-free webp encoder/decoder: git.sr.ht/~jackmordaunt/go-libwebp :)

laszlo1337:
Hello. I understand there are difficulties with this issue, but since Safari now supports webp (finally after 13 years smh) I anticipate real uptick in WEBP format adoption, since it's a superior format in many aspects. So I am just asking if perhaps including this improvement could be moved up the schedule - especially if, like @Wikidepia mentioned, a solution is available. Thank you

ganigeorgiev:
@laszlo1337 I'm not planning working on this for now as other tasks are with higher priority (see https://github.com/pocketbase/pocketbase/discussions/4355). I haven't tried the suggested library and I'm not sure how difficult it would be to integrate with, nor what platforms are supported (_see also this comment of the author of the above library - https://github.com/golang/go/issues/45121#issuecomment-1880348171_).
[...]

PR #6744, opened 2025-04-19
Generate webp thumbnails

KevSlashNull:
This adds support for webp thumbnails by explicitly importing the
official (but experimental) [x/image/webp](https://pkg.go.dev/golang.org/x/image/webp) decoder for webp.

Unless the decoder panics (can it?), this is safe. Pocketbase will fall
back to reusing the fullsize image if creating a thumbnail fails. Which
is already the current behavior for all webp images.

Since disintegration/imaging is not aware of webp, webp thumbnails will
be generated in png format. This is a caveat but modern web browsers
generally don’t rely on file extensions to detect image formats.

### DIFF
--- a/apis/file.go
+++ b/apis/file.go
@@ -19,7 +19,7 @@ import (
 	"golang.org/x/sync/singleflight"
 )
 
-var imageContentTypes = []string{"image/png", "image/jpg", "image/jpeg", "image/gif"}
+var imageContentTypes = []string{"image/png", "image/jpg", "image/jpeg", "image/gif", "image/webp"}
 var defaultThumbSizes = []string{"100x100"}
 
 // bindFileApi registers the file api endpoints and the corresponding handlers.
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/spf13/cast v1.7.1
 	github.com/spf13/cobra v1.9.1
 	golang.org/x/crypto v0.37.0
+	golang.org/x/image v0.26.0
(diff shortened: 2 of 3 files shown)


KevSlashNull:
hey @ganigeorgiev :wave: WDYT about this?

i saw that you wrote in https://github.com/pocketbase/pocketbase/issues/825#issuecomment-1288003603 that webp thumbs would be out-of-scope for v1.0, so no worries if you don’t want to include this. i just thought it would be neat because i use webp images a lot!

ganigeorgiev:
I'm not really sure. As mentioned in the linked comment, I wanted to avoid the workaround of converting them to jpg/png but since webp thumbs seems to be a common request I guess it could be better than nothing. I'll have to think a little more on it.

ganigeorgiev:
In the end it is better than nothing, so let's merge it. If it ended up causing issues we can always revert it.

I'll update also the tests to add a mimetype thumb check to ensure that the fallback png is properly generated.

qoheleth-tech:
> webp thumbnails will be generated in png format


Does this mean that the thumbnail will be a `webp` file but have a `png` extension; or that the thumbnail will be a `png` with a `png` extension; or that the  thumbnail will keep the `webp` extension but actually be a `png` image? Thanks for your work on this!

ganigeorgiev:
@qoheleth-tech The generated thumb image should be png but the extension will be the same as the original (it shouldn't really matter anyway because we don't trust it).

PR #6861, opened 2025-05-20
Add full support for WebP

gocmarek:
As he subject of this PR. This make the full support for WebP images - Decode/Encode


ganigeorgiev:
I think I've been clear in your earlier opened issue that I don't want this. Please don't spam.

PR #6879, opened 2025-05-25
Disable TLS when `--https=""`

xg1990:
According the argument description:

```
TCP address to listen for the HTTPS server
                          (if domain args are specified - default to 0.0.0.0:443, otherwise - default to empty string, aka. no TLS)
                          The incoming HTTP traffic also will be auto redirected to the HTTPS version
```

However, if set `--https=""`, the current code will force enabling TLS on `0.0.0.0:443`, which is different from the described behavior.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -27,9 +27,6 @@ func NewServeCommand(app core.App, showStartBanner bool) *cobra.Command {
 				if httpAddr == "" {
 					httpAddr = "0.0.0.0:80"
 				}
-				if httpsAddr == "" {
-					httpsAddr = "0.0.0.0:443"
-				}
 			} else {
 				if httpAddr == "" {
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
No it works correctly.

If at least one domain is specified then the default http and https listeners should remain.

Feel free to elaborate what actually is the issue you are trying to solve but I don't think the current behavior needs to change.

xg1990:
> No it works correctly.
> 
> 
> 
> If at least one domain is specified then the default http and https listeners should remain.
> 
> 
> 
> Feel free to elaborate what actually is the issue you are trying to solve but I don't think the current behavior needs to change. 

I was trying to start the server and listen to 0.0.0.0:80 only, with https disabled. Is there any way to achieve that?

PR #6914, opened 2025-06-08
S3 Storage use default "bytes=0-" range instead of "" empty range

joennlae:
First of all. What a fantastic project :-). You are an absolute beast.

## Changes

This PR changes from `Range` header equals `""` to `bytes=0-` for S3 storage driver.

## Backstory

I ran into an error with our S3 Storage provider (https://flow.swiss/object-storage, we have to use it due to data privacy reasons...). I assume they use `minio` hidden behind a `caddy` reverse proxy. 

This is the error I got:
```
416 InvalidRange: The requested range cannot be satisfied.
(RAW: <?xml version="1.0" encoding="UTF-8"?><Error><Code>InvalidRange</Code><Message>The requested range cannot be satisfied.</Message></Error>)
```

I wrote some local tests scripts. Range is fully supported by the provider. 

Now in the case where pocketbase requests the full file.

https://github.com/pocketbase/pocketbase/blob/88bb8c406e6cb49b81f57007b8511ccdccf80610/tools/filesystem/blob/bucket.go#L409-L411

The logic triggered results in a `Range` header equals `""`.

https://github.com/pocketbase/pocketbase/blob/88bb8c406e6cb49b81f57007b8511ccdccf80610/tools/filesystem/internal/s3blob/s3blob.go#L192-L200

In the `minio` test we can see the unsupported `range` equals `""` here:

https://github.com/minio/minio/blob/21409f112dc299966a3beca89ead13f5045ecc33/cmd/httprange_test.go#L62

### DIFF
--- a/tools/filesystem/internal/s3blob/s3blob.go
+++ b/tools/filesystem/internal/s3blob/s3blob.go
@@ -199,8 +199,18 @@ func (drv *driver) NewRangeReader(ctx context.Context, key string, offset, lengt
 		byteRange = fmt.Sprintf("bytes=%d-%d", offset, offset+length-1)
 	}
 
-	reqOpt := func(req *http.Request) {
-		req.Header.Set("Range", byteRange)
+	var reqOpt func(*http.Request)
+
+	if byteRange != "" {
+		reqOpt = func(req *http.Request) {
+			req.Header.Set("Range", byteRange)
+		}
+	} else {
+		// Always set a range header, even for full reads
+		byteRange = "bytes=0-"
+		reqOpt = func(req *http.Request) {
+			req.Header.Set("Range", byteRange)
+		}
 	}
 
 	resp, err := drv.s3.GetObject(ctx, key, reqOpt)


ganigeorgiev:
If the empty string value is causing compatibility issues, can we just skip the `Range` header in that case?

Could you test changing `reqOpt` to something like:

```go
reqOpt := func(req *http.Request) {
    if byteRange != "" {
        req.Header.Set("Range", byteRange)
    }
}
```
?

I'll investigate it in more details tomorrow after work.

joennlae:
That was my first idea. When using this there is no S3 error but the file is empty (having zero bytes). That is why I opted for the `bytes=0-` solution.

ganigeorgiev:
I'm confused but also too tired to debug it at the moment.

At quick glance over the suggested change, if `bytes=0-` is really necessary then assigning it in the `else` clause of the original checks would be better.

But I don't think that the `Range` header is required by the `GetObject` S3 spec and it should work without it.

I also remember testing PocketBase with the Minio demo playground and it was working fine as it is with the empty header so it is possible to be something Flow specific (_or the playground to had a proxy that excludes empty headers?_).

joennlae:
Just that there is no confusion. I assume the flow.swiss is using minio. Looking more into it: it is most probable that they have some misconfiguration or outdated version.

I can also close this and just patch it on my side. I unfortunately do not have the time to make a full detailed investigation.

I can request the version of the service they run from flow.swiss tomorrow.

joennlae:
My next steps would be:
- Write flow.swiss which S3 compatible storage they are using
- try to make a minmal reproducible setup (if even possible)

ganigeorgiev:
I've asked Flow to grant me free trial without verification and they agreed so I'll be able to test the issue myself later today.

I'm not against adding `bytes=0-` as fallback value but the header should be optional and I wanted to verify it on my own whether it is really necessary or not.

joennlae:
I investigated a little bit more because I completely agree with your train of thought.

If I do not set the "Range" header (which I also think is the better solution). I get to complete answer (in python (boto3 and plain html) and also in pocketbase):

```
S3 GetObject Request Headers:
---
S3 GetObject Response Headers:
  Date: [Mon, 09 Jun 2025 08:13:05 GMT]
  Etag: [W/"8bfa8e0684108f419933a5995264d150"]
  Server: [Caddy nginx]
  X-Amz-Meta-Original-Filename: [test.txt]
  Content-Type: [text/plain; charset=utf-8]
  X-Amz-Req-Time-Micros: [5592]
  X-Amz-Request-Id: [80000000000000b800047e7642bd17c2]
  X-Amz-Storage-Class: [STANDARD]
  Last-Modified: [Mon, 09 Jun 2025 08:13:05 GMT]
  Vary: [Accept-Encoding]
---
S3 GetObject Parsed Response:
  ContentType: text/plain; charset=utf-8
  ContentLength: 0 bytes
  ETag: W/"8bfa8e0684108f419933a5995264d150"
  LastModified: 2025-06-09 08:13:05 +0000 GMT
  ContentRange: 
  ContentEncoding: 
  CacheControl: 
  ContentDisposition: 
  ContentLanguage: 
  Metadata: map[original-filename:test.txt]
---
S3 GetObject Response Body (first 12 bytes):
Test content
```

The Content-Length is parsed to 0.

The `ContentLength = 0` is then consumed here: https://github.com/pocketbase/pocketbase/blob/88bb8c406e6cb49b81f57007b8511ccdccf80610/tools/filesystem/internal/s3blob/s3blob.go#L216-L224

I tried to investigate it even more:
- When I use a Pre-Signed URL like this: https://os.alp1.flow.swiss/data2u-data-test/pbc_2661488127/02dq9w752o2lux9/test_ilim9k3hsq.txt?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=8d50e20c55ea9ea4EDVI%2F20250609%2Falp1%2Fs3%2Faws4_request&X-Amz-Date=20250609T083452Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=c032990b1c863f464fa81887c301bf7b0bc24386b0bfa1bf1bf5b7478126c835

I get the `Content-Length`. 

When calling from pocketbase it is empty. I have a hunch (but it is unconfirmed).

https://github.com/pocketbase/pocketbase/blob/88bb8c406e6cb49b81f57007b8511ccdccf80610/tools/filesystem/internal/s3blob/s3/s3.go#L132

ganigeorgiev:
I've just tested it locally against a Flow bucket in ZRH1 region and it works correctly for me when skipping the range header. `Content-Length` is also returned and non-empty.

I don't know why it doesn't work in your case but I suspect that the `Content-Length` header is being stripped by the proxies or it is possible to be some obscure bug in Flow.

In any case, although other providers just ignore it, we shouldn't send `Range: ` header if there is no bytes range requested since the HTTP spec doesn't seem to allow empty string as `byte-ranges-specifier`.

The change currently is in `master` and will be available with the next minor release a little bit later today.

If you still think that there is a bug somewhere with the PocketBase S3 implementation feel free to elaborate and I'll to investigate it further, but again it works correctly for me when testing locally against the free trial Flow Object storage instance.

LeCyreaxYT:
Hey, @ganigeorgiev 

I have testet it with the Object Storage Location at ALP1
and my results are same like from @joennlae 

Error:
`[key: _pb_users_auth_/ukjdyqlxadiz904/thumbs_screenshot_2025_01_14_220235_ucxptjq7jd.png/100x100_screenshot_2025_01_14_220235_ucxptjq7jd.png] 416 InvalidRange: The requested range cannot be satisfied.
(RAW: <?xml version="1.0" encoding="UTF-8"?><Error><Code>InvalidRange</Code><Message>The requested range cannot be satisfied.</Message></Error>)`

ganigeorgiev:
@LeCyreaxYT Did you test with the change in `master`?

LeCyreaxYT:
Not jet but i will test it.

ganigeorgiev:
@LeCyreaxYT The `The requested range cannot be satisfied` error is returned because we are sending empty `Range: ` header but as mentioned in my previous reply skipping the header when `byteRange` is empty works correctly for me.

The issue @joennlae has is that `Content-Length` is returned with 0 bytes for some unknown reason.

joennlae:
> I've just tested it locally against a Flow bucket in ZRH1 region and it works correctly for me when skipping the range header. `Content-Length` is also returned and non-empty.
> 
> I don't know why it doesn't work in your case but I suspect that the `Content-Length` header is being stripped by the proxies or it is possible to be some obscure bug in Flow.
> 
> In any case, although other providers just ignore it, we shouldn't send `Range: ` header if there is no bytes range requested since the HTTP spec doesn't seem to allow empty string as `byte-ranges-specifier`.
> 
> The change currently is in `master` and will be available with the next minor release a little bit later today.
> 
> If you still think that there is a bug somewhere with the PocketBase S3 implementation feel free to elaborate and I'll to investigate it further, but again it works correctly for me when testing locally against the free trial Flow Object storage instance.

Thank you for all your effort. I have the same conclusion as you have drawn.

joennlae:
I quickly tested the master with ALP1. It still has the issue that `Content-Length` is returned as `0`. So no error is thrown in pocketbase but the file that is returned has zero bytes in it.

I will have the local workaround for now.

joennlae:
I quickly tested it on `ZRH1` and the master works...
So it is definitely a `flow` issue.

ganigeorgiev:
_Just for info in case someone stumble here - Flow contacted me regarding the missing `Content-Length` and told me that they've updated their software in the ALP1 region; I've briefly tested again with all 3 regions and it works correctly for me. If someone is still experiencing the same issue, please report it to the Flow support._

PR #6940, opened 2025-06-18
Fix typo in ApiScenario

davenh99:
Hi Gani,

I noticed this double quote was missing when writing a test.


ganigeorgiev:
Thank you, I've merged it locally in `develop`.

PR #6991, opened 2025-07-05
feat: Minimal PocketBase with Google OAuth only

ghost:
- Disabled Admin UI by commenting out its route and startup banner link.
- Removed email/password authentication routes from API.
- Removed email/password related email sending functions.

This results in a PocketBase backend focused solely on API delivery with Google OAuth as the exclusive authentication method.


ganigeorgiev:
I don't understand what is the purpose of this pull request but please don't spam and read https://github.com/pocketbase/pocketbase?tab=readme-ov-file#contributing.

If all you want to disable some of the builtin routes then you can register a middleware that checks for their paths and return an error.

PR #6996, opened 2025-07-07
add exit code

nathancahill:
Noticed this when superuser commands throw errors, the exit code is always 0. This sets the exit code to 1 if an error is returned.

### DIFF
--- a/pocketbase.go
+++ b/pocketbase.go
@@ -167,7 +167,10 @@ func (pb *PocketBase) Start() error {
 	pb.RootCmd.AddCommand(cmd.NewSuperuserCommand(pb))
 	pb.RootCmd.AddCommand(cmd.NewServeCommand(pb, !pb.hideStartBanner))
 
-	return pb.Execute()
+	err := pb.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
 }
 
 // Execute initializes the application (if not already) and executes


ganigeorgiev:
This is similar to https://github.com/pocketbase/pocketbase/pull/5399.

For more details, see also https://github.com/pocketbase/pocketbase/discussions/2738.

nathancahill:
Hard to integrate PB with other tools if it doesn't follow standard practices. We'll have in inspect the output for `error:`

ganigeorgiev:
Again, if we kill the program with `os.Exit` we can't gracefully shutdown the application and this in my opinion is more important, especially when this also affects the proper database connections termination.

There are option to "lazily" exit once the the graceful shutdown handling completes but it complicates too many things for very little benefit and for now it is left aside.

If you have an idea how this could be better handled, feel free to create a proposal and I'll try to at least have a look at it.

nathancahill:
I'm not so familiar with the codebase, but wouldn't any situations where we're returning an error from pb.Execute not need a graceful shutdown? Are there any specific code paths where an error is returned that need cleanup? With a bit more information I can try to propose a solution.

ganigeorgiev:
> ...where an error is returned that need cleanup? 

Any error could require cleanup. That's the difficult part because it is hard to generalize.

Someone could have for example a temp file creation and would expect their `defer` call to execute and delete the file but killing the program with `os.Exit` will prevent that.

Is `os.Exit` really that important for your use case?

nathancahill:
From the `superuser` subcommand, none of the errors require cleanup.

From the `serve` subcommand, we have 5 places we're returning errors:

1. `err := app.RunAllMigrations()` (fine to exit without cleanup)
2. `err := NewRouter(app)` (fine to exit without cleanup)
3. `err := app.OnServe()...` (fine to exit without cleanup)
    a. err := e.Router.BuildMux() 
    b. err := net.Listen()
4. `if listener == nil` (fine to exit without cleanup)
5. `err := serveEvent.Server.Serve/ServeTLS` __<- (this should have cleanup)__

I'd propose to mark the serveEvent error as a "soft error" that allows cleanup, handle everything else with `os.Exit`. So returning `&SoftError{Err: serveErr}` from [here](https://github.com/pocketbase/pocketbase/blob/master/apis/serve.go#L293) and [here](https://github.com/pocketbase/pocketbase/blob/master/apis/serve.go#L296) and then handling these from the command:

```
func (pb *PocketBase) Start() error {
	// register system commands
	pb.RootCmd.AddCommand(cmd.NewSuperuserCommand(pb))
	pb.RootCmd.AddCommand(cmd.NewServeCommand(pb, !pb.hideStartBanner))

	err := pb.Execute()
	if err != nil && !errors.As(err, &softErr) {
		os.Exit(1)
	}
}
```

ganigeorgiev:
Returning a custom error was one of the options that was considered in the past but again just because a command doesn't have a cleanup procedure that doesn't mean that there is no other code (e.g. app hooks) that may require it.

There is no guarantee of course and your app can be always killed abruptly by the OS, but once the application is bootstrapped it is usually expected that a graceful shutdown handling will occur in normal operation _(as a minimum there are users that actively rely on the `app.OnTerminate` hook)_.

I agree that returning a proper exit code would be nice, but the scenarios where you'll need some external application to know with what exit code your PocketBase process ended is very rare in my experience, and the current lack of non-0 exit code for the commands error is simply the "safer" option.

Other users suggested in the past to make use of `runtime.Goexit()` in the main goroutine and it is something that exists in my todo to experiment with it, but for now remains a very low priority unless a more prominent use case arise.

nathancahill:
The use case is creating superusers programmatically before the server is started. Let's set aside the whole graceful shutdown aspect, and instead start by returning proper error codes from the superuser subcommand, leaving the serve command functionality as is. If the superuser subcommand errors, we can call `os.Exit(1)`. Is that a good intermediary solution?

ganigeorgiev:
>  If the superuser subcommand errors, we can call os.Exit(1). Is that a good intermediary solution?

No, because again at minimum `app.OnTerminate` will not be triggered.

ganigeorgiev:
_Just for info in case someone stumble here: non-zero exit code support for commands returning an error (or panic) was added in the `develop` branch and will be available with the upcoming PocketBase v0.40.0 release _(there are no ETAs)_._

PR #7005, opened 2025-07-11
clear sass warning

yeclimeric:


yeclimeric:
merge dev@ui

ganigeorgiev:
Thank you but this is not necessary as there are plans to rewrite the UI entirely (and eventually even remove Sass).

yeclimeric:
Thank you very much for your reply. We look forward to the new UI.

Issue #7035, opened 2025-07-22
Request for X Oauth implementation


ankushKun:
I am trying to implement X Oauth in my application through pocketbase and I noticed that pocketbase still sends api calls to twitter.com instead of x.com for oauth request, causing the user to login twice on twitter.com, even though they might already be logged in on x.com

Moreover, the twitter api does not respond with an email address for the auth user, causing pocketbase user record creation to fail (note: 'Request email from users' is enabled in my X app auth settings)

```
{
    "data": {
        "email": {
            "code": "validation_required",
            "message": "Cannot be blank."
        }
    },
    "message": "Failed to create record.",
    "status": 400
}
```

ankushKun:
I can disable the non empty toggle to false for the email field as a temporary fix, so as to not get any errors, but then if the user authenticates with a different provider, it creates a new user instead of adding an authorised provider to the existing user.

ganigeorgiev:
In the past the Twitter OAuth2 didn't support returning an email address. I'm not sure what is the state of the APIs at the moment as I haven't used Twitter/X in a very long time.

I'll try to find time and research it in more details eventually this weekend, but user contributions are welcomed for this. The provider could be found in https://github.com/pocketbase/pocketbase/blob/master/tools/auth/twitter.go. For backward-compatibility, we can also eventually create a new provider specifically for "X".

ganigeorgiev:
I've pushed an update in the `develop` branch that uses the `x.com` domain and should return the `confirmed_email` user field. The provider name is kept `twitter` to avoid introducing unnecessary breaking changes for existing users.

The fix will be available with the next minor v0.29.1 release _(most likely next weekend)_.

ankushKun:
Tried it just now, works perfect!

Thank you :)

PR #7043, opened 2025-07-23
created oauth for X

iamritikbhardwaj:
Hi community i tried making this oauth to fix the twitter issue #7035
- I am fairly new to this
- Please help me figure out how can i test this
- Any feedbacks will be appriciated

### DIFF
--- a/tools/auth/x.go
+++ b/tools/auth/x.go
@@ -1,0 +1,72 @@
+package auth
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/pocketbase/pocketbase/tools/types"
+	"golang.org/x/oauth2"
+)
+
+type X struct {
+	BaseProvider
+}
+
+const NameX string = "X"
+
+func init() {
+	Providers[NameX] = wrapFactory(NewXProvider)
+}
+
+func NewXProvider() *X {
+	return &X{BaseProvider{
+		ctx:         context.Background(),
+		displayName: "X",
+		pkce:        true,
+		scopes: []string{
+			"users.read",
+			"tweet.read",
+		},
+		authURL:     "https://x.com/i/oauth2/authorize",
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Did you actually test it at least locally with a static HTML?

The `email` field doesn't seem mapped anywhere and in addition we need information that the email can be trusted, aka. it is verified.

The default scopes also may need to change.

The issue is not just swapping "twitter" with "x" domain. Please test it and provide links to the resources that you used to find the necessary information.

ganigeorgiev:
Thank you for the effort but it seems that you haven't researched it. I'll investigate it on my own this weekend.

iamritikbhardwaj:
> Thank you for the effort but it doesn't seem that you've researched it. I'll investigate it on my own this weekend.

https://pkg.go.dev/golang.org/x/oauth2/endpoints?utm_source=chatgpt.com
https://docs.x.com/resources/fundamentals/authentication/oauth-2-0/user-access-token

These are the resources i researched from and i appriciate for your feedback so about static page how do i test this and can you please provide a detailed feedback if it's not a hassle.
- Do let me know if i am being annoying but i am curious and have some questions
I have not added any email so please let me know where is it

ganigeorgiev:
I'm sorry but the lack of proper research defeats the purpose of the contribution and I don't consider it helpful. I can change the domain name on my own if that was the only issue. The main problem is that we need to find out whether the `email` can be returned and whether there is need for an extra scope. Blindly applying something that ChatGPT suggested is not welcomed here.

iamritikbhardwaj:
Okay i will keep that in mind just let me know how can i help i am trying to learn things and i assure you i use chatgpt for research only all the code you see is written by me only.
My only purpose is to learn from professional thank you
And if you are kind enough to point me i right direction maybe some other issue do work on ill gladly do the research .

PR #7056, opened 2025-07-29
add Box.com as oauth2 provider. Details :

blakepatteson:
1.  implement Box OAuth2 authentication provider, following pattern of all the others
2.  add Box provider configuration to UI
3.  include Box logo .svg file and tests
4.  update provider count in tests

I built / tested the front-end after running `npm run build` but deleted the js bundles so as to keep the diff clean... I'll happily rebuild it if requested.

  API reference links :
      https://developer.box.com/reference/get-users-me/
      https://developer.box.com/guides/authentication/oauth2/oauth2-setup/


tools/auth/box.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if err := json.Unmarshal(data, &extracted); err != nil {
+		return nil, err
+	}
+
+	user := &AuthUser{
+		Id:           extracted.ID,
+		Name:         extracted.Name,
+		Username:     extracted.Login,
+		Email:        extracted.Login, // Box uses login as email identifier
```
ganigeorgiev, line 91: Please double check that the `extracted.Login` field is a valid email address and that it is always verified.

tools/auth/box.go
```diff
@@ -0,0 +1,101 @@
...
+	}{}
+
+	if err := json.Unmarshal(data, &extracted); err != nil {
+		return nil, err
+	}
+
+	user := &AuthUser{
+		Id:           extracted.ID,
+		Name:         extracted.Name,
+		Username:     extracted.Login,
```
ganigeorgiev, line 90: If the provider doesn't return a username, do not assign it to the email and leave it empty instead so that it can be autogenerated.

ganigeorgiev (changes requested):
Thank you, I'll have a more detailed look at it eventually this weekend _(or the next one)_.

For now I left some nitpicks that you could check when you find time.

blakepatteson:
Thanks so much for the feedback! 

I will make these updates shortly and mark the pull request as "Ready for Review". 

PS : thank you for putting PocketBase out there - it's an amazing tool - so simple, but still useful in our times of increasingly complicated software. and thanks for making it easy to contribute!

ganigeorgiev:
Thank you for verifying and applying the review notes. I've also tested the provider with a trial account and everything seems to work fine.

I've merged the changes locally for now in the `develop` branch _(with some minor adjustments to the logo to be more close to the official [Box branding style guide](https://cloud.app.box.com/v/externalstyleguide))_.

The Box provider will be part of the next patch release together with some other OAuth2 Twitter changes _(most likely sometime later today)_.

PR #7116, opened 2025-08-20
Separate syscall import with build constraints (WASM)

joas8211:
Separating syscall import with build constraints allows Pocketbase core to be used for building to WASM. Compiling WASM build does not allow that `import "syscall"` appears anywhere within used source code files. This was the only real blocker for building to WASM that cannot be accounted for in the module consuming Pocketbase. See https://github.com/joas8211/pocketbase-wasm for the proof of concept.

Without these changes, the following error is logged on build:

```
# github.com/pocketbase/pocketbase/core
pocketbase/core/base.go:776:18: undefined: syscall.Exec
```

### DIFF
--- a/core/base.go
+++ b/core/base.go
@@ -12,7 +12,6 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/fatih/color"
@@ -773,7 +772,7 @@ func (app *BaseApp) Restart() error {
 			}
 		}()
 
-		return syscall.Exec(execPath, os.Args, os.Environ())
+		return exec(execPath, os.Args, os.Environ())
 	})
 }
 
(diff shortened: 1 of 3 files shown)


ganigeorgiev:
Thank you. I'm not sure how useful WASM build will be in practice but it is definitely an interesting experiment and I'm curious to see how far you'll take it. I'll check the PR in more details this weekend and eventually include it in the planned patch release.

> Compiling WASM build does not allow that import "syscall" appears anywhere within used source code files.

I don't think it is the `syscall` package import as a whole that it is causing the issue and it is most likely because of the `Exec` call. There is [`syscall` for `GOOS=js`](https://pkg.go.dev/syscall?GOOS=js) but for some reason the `Exec` function is not defined _(Windows also doesn't implement it but at least it is defined and an error is returned for this case - [`exec_windows.go`](https://cs.opensource.google/go/go/+/refs/tags/go1.25.0:src/syscall/exec_windows.go;l=403))_.

_Side-note: you can exclude the default SQLite driver with the `no_default_driver` build tag._

PR #7135, opened 2025-08-26
feat(schema): Merge scaffold indexes on collection type change

PatschD:
Currently there is a bug when creating a new auth collection. The indexes from the scaffold are not correctly set. In the image below you should see 2 indexes but it shows as 0:
image

this also leads to missing fields in the option tab (since these fields are derived from unique indexes):
image

however when you submit the collection without changing anything the indexes do get set somewhere: 
image

i think the problem happens here: 

```
CollectionUpsertPanel.svelte line 280:

collection = Object.assign(structuredClone($scaffolds[t]), collection);
``` 

you overwrite the fields in ```$scaffolds``` with the fields from collection (which has 3 fields (id, updated, created), and index:[]), however because you bind the collection:                
 
```
CollectionUpsertPanel.svelte line 548:
<CollectionFieldsTab bind:collection />
 ```

the fields that are required from the auth collection are added in ```onTypeChange```, however the indexes required for auth are not. i added the code to merge them too. I think indexes are the only property that need to be added manully like this. 

This change adds and displays the indexes properly:

image

### DIFF
--- a/ui/src/components/collections/CollectionFieldsTab.svelte
+++ b/ui/src/components/collections/CollectionFieldsTab.svelte
@@ -145,6 +145,12 @@
         const blank = structuredClone($scaffolds[collection.type]);
         collection.fields = blank.fields;
 
+        // Merge existing indexes with the new type's scaffold defaults.
+        // Using a Set is a clean way to combine the arrays and prevent duplicates.
+        const oldIndexes = collection.indexes || [];
+        const newIndexes = blank.indexes || [];
+        collection.indexes = [...new Set([...oldIndexes, ...newIndexes])];
+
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Thank you. I've pushed a slightly different implementation to account for the case when switching back-and-forth between the collection types to minimize potential validation errors.

Technically we can also just "reset" the indexes with the new scaffold ones but since we already attempt to merge the fields we can keep it.

There are some other minor things related to the scaffolds that I've noticed and I'll push a little bit later.

For now the "fix" is merged in the `develop` branch and eventually will be available with the next minor release (most likely ~this~ next weekend).

PR #7136, opened 2025-08-26
Refactor to use reflect.TypeFor

cuiweixie:


ganigeorgiev:
Thank you. I've added a note to consider replacing it eventually with the next release but for now it is not really necessary.

PR #7196, opened 2025-09-21
Add configurable max render length for Text columns in Admin UI

TimeLord2010:
This PR adds an optional "Max render length" setting when creating Text columns, allowing users to control how many characters are displayed before truncation in the webadmin table view.

Changes:
- New optional text field in column configuration (defaults to 150 chars if empty/zero)
- Text now wraps within the specified length before applying ellipsis
- Previously: no wrapping, ellipsis at column width
- Now: wrapping enabled, ellipsis at configured character count

Context:
I understand this is primarily a webadmin UI enhancement without direct API benefits, so I completely understand if this doesn't align with the project's direction. However, it solves a specific need in my PocketBase implementation, and I wanted to contribute it back in case others find it useful.

Happy to make any adjustments or close this if it's not a good fit for the core project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 
 .DS_Store
 
+/ui/dist/
+CLAUDE.md
+
 # goreleaser builds folder
 /.builds/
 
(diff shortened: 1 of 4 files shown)


ganigeorgiev:
I don't think this needs to be configurable and please don't open AI PRs.

Issue #35, opened 2022-07-09
How to add  AUTOINCREMENT int field?


sospartan:
The `id` field is string and the number type don't have a `autoincrement` option.
Is there any way to achieve it?

sospartan:
I'll use `rowid` for my purpose. Close this one.

PR #7198, opened 2025-09-22
Fix DateTime field empty value insert failure in PostgreSQL #35

fondoger:


ganigeorgiev:
If you are going to spam me even though I mentioned clearly that I'm not interested in merging or adding PostgreSQL support, then I don't want to collaborate with you anymore. Please just use something else.

PR #7208, opened 2025-09-27
Exclude the lost+found folder from backups

lbndev:
Hello,

When the pb_data folder is mounted from another filesystem, it will contain a "lost+found" folder which is accessible only to root.
If pocketbase is running as a non-root user, the backup process fails because this folder can't be read.

I have added "lost+found" to the list of excluded folders in the backup logic.

### DIFF
--- a/core/base.go
+++ b/core/base.go
@@ -40,6 +40,7 @@ const (
 	LocalBackupsDirName       string = "backups"
 	LocalTempDirName          string = ".pb_temp_to_delete" // temp pb_data sub directory that will be deleted on each app.Bootstrap()
 	LocalAutocertCacheDirName string = ".autocert_cache"
+	LocalLostFoundDirName     string = "lost+found"
 )
 
 // FilesManager defines an interface with common methods that files manager models should implement.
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
I'll have to read a little bit more about that but in the meantime, could you provide a reproducible example/steps for the claim:

> When the pb_data folder is mounted from another filesystem, it will contain a "lost+found" folder which is accessible only to root.

?

ganigeorgiev:
OK, after [quick read](https://tldp.org/LDP/Linux-Filesystem-Hierarchy/html/lostfound.html) about it, this directory seems usually created and managed by `fsck` and I assume that you are seeing it in your `pb_data` because you've mounted it as the root destination for your volume/filesystem. If that's the case, then Yes, it will require elevated permissions to perform the backup.

I guess it won't hurt having it excluded by default so I'll consider merging it tomorrow (and maybe create a dummy `lost+found` dir in the `test/data` for the backup tests).

ganigeorgiev:
I've merged it locally in the `develop` branch and it will be available with the next minor release (most likely sometime next week).

_Side-note: there are plans to optimize and refactor the backups in order to minimize the locking times when you have a lot of uploaded files, and the backup exclusion filter most likely will be reverted to be "included only" per several other user complains._

PR #7218, opened 2025-09-30
Fix incomplete multipart uploads when MaxConcurrency is 1

arafat4693:
When `MaxConcurrency` is set to 1 and a multipart upload is initiated with an `initPart`, the code decrements `totalParallel` from 1 to 0, resulting in zero worker goroutines being started for the remaining file parts. This causes only the first part to be uploaded while the rest of the file is silently skipped.

This fix ensures at least one worker goroutine is always started to process the remaining parts, regardless of the `MaxConcurrency` value.

Fixes issue where files larger than `MinPartSize` would be incompletely uploaded when `MaxConcurrency=1`.

### DIFF
--- a/tools/filesystem/internal/s3blob/s3/uploader.go
+++ b/tools/filesystem/internal/s3blob/s3/uploader.go
@@ -353,6 +353,11 @@ func (u *Uploader) multipartUpload(ctx context.Context, initPart []byte, optReqF
 		})
 	}
 
+	// Ensure that the totalParallel is at least 1
+	if totalParallel < 1 {
+		totalParallel = 1
+	}
+
 	for i := 0; i < totalParallel; i++ {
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Thank you. I've added a note to investigate it this weekend and eventually include it in the planned bugfix release, but for now I don't want to rush merging it since this functionality is not exposed to the filesystem `Upload*` methods and the tests also will have to be revisited.

ganigeorgiev:
I've pushed a slightly different implementation and updated the tests, but again this functionality is not exposed at the moment though the filesystem `Upload*` methods and shouldn't really have any effect on normal applications usage, so if the above report was generated via some AI tool, please don't do that again and instead elaborate more on your actual usage.

PR #7255, opened 2025-10-15
fix(s3blob): Correct file size detection for R2-like providers

GetterSethya:
Some S3-compatible object storage providers, like Cloudflare R2, may not consistently return the Content-Length header for GET requests on certain file types (e.g., non-image files). This can lead to PocketBase incorrectly identifying a file's size as 0.

This pull request resolves the issue by implementing a fallback mechanism. If a GET request for a full object results in a size of 0, the code now performs an additional HEAD request to reliably retrieve the correct content length.


ganigeorgiev:
No, I don't think this is a good idea.

If Cloudflare R2 doesn't return the content length for whatever reason, then this should be reported to the upstream in case it is a bug and need fixing, and not workaround it by sending an extra request.

PR #7262, opened 2025-10-18
Protect changing in parallel using HTTP 412 precondition failed

tbalsys:
Screenshot 2025-10-18 at 17 03 58


ganigeorgiev:
Hey, please don't open PR like that!

tbalsys:
Hi, would you be so kind to hint how to open the PR? Should I open an issue first? Wrong branch name? Is code not correct or feature undesirable?

ganigeorgiev:
First there is no information why this is needed in the first place.

Second, this feature is not desired - not only it will be a breaking change, the implementation is also not OK (there is no need to be part of the core package if you already using the hooks). The UI is also in the process of being refactored and realtime subscriptions are planned to be enabled so if the goal is to prevent accidentally submitting an outdated change, then we can consider eventually adding a warning before save instead of rejecting the request.

And third, please read the contribution note in the README and more specifically:
> But please refrain creating PRs for new features without previously discussing the implementation details. PocketBase has a [roadmap](https://github.com/orgs/pocketbase/projects/2) and I try to work on issues in specific order and such PRs often come in out of nowhere and skew all initial planning with tedious back-and-forth communication.

So as kindly as possible, your contribution is not welcomed especially in this form.

PR #7294, opened 2025-11-02
Replace PseudorandomStringWithAlphabet to use math/rand/v2

corvad:
Just wanted to clean up a todo and updated the math/rand to math/rand/v2. Simple change and all tests seem to still pass.

David Corvaglia


ganigeorgiev:
Thank you but it is not really necessary right now.

PR #7305, opened 2025-11-07
Fix deadlock bug when triggering a termination

yerTools:
Hey,

I'm pretty sure I found a deadlock bug when triggering a termination.

## Context

I wanted a graceful shutdown option for an app I'm working on.
Thats why I took a look how you implemented it.

PocketBase uses an interrupt signal (`signal.Notify(sigch, os.Interrupt, syscall.SIGTERM)`) which triggers the termination event through the `done` channel (from `/pocketbase.go`):

```go
done := make(chan bool, 1)

// listen for interrupt signal to gracefully shutdown the application
go func() {
    sigch := make(chan os.Signal, 1)
    signal.Notify(sigch, os.Interrupt, syscall.SIGTERM)
    <-sigch

    done <- true
}()

// execute the root command
go func() {
    // note: leave to the commands to decide whether to print their error
    pb.RootCmd.Execute()

    done <- true
}()

<-done

// trigger cleanups
event := new(core.TerminateEvent)
event.App = pb
return pb.OnTerminate().Trigger(event, func(e *core.TerminateEvent) error {
    return e.App.ResetBootstrapState()
})
```

So I copied the code for triggering a termination event into my project (`app.OnTerminate().Trigger(&core.TerminateEvent{App: app})`).

But that triggered a deadlock.

## How this can happen

When the application runs normally and the termination event fires, the registered handlers are executed and the web server is shut down as intended.

In the snippet above, `pb.RootCmd.Execute()` starts the web server, so once the server shuts down the function returns, writes `true` into `done`, and PocketBase - still waiting on `<-done` - fires another termination event.

**Up to this point everything is fine**: the termination handlers run a second time, which should be harmless for all handlers except one.

## The problematic handler

The handler in question is (`__pbAppLoggerOnTerminate__`) in `/core/base.go`:

```go
app.OnTerminate().Bind(&hook.Handler[*TerminateEvent]{
    Id: "__pbAppLoggerOnTerminate__",
    Func: func(e *TerminateEvent) error {
        handler.WriteAll(context.Background())

        ticker.Stop()

        done <- true

        return e.Next()
    },
    Priority: -999,
})
```

The `done` channel is unbuffered, and the real issue is `done <- true`. Adding a buffer helps, but if the handler runs `x` times you’d need a buffer of size `x - 1`, which is arbitrary.

That's why I replaced the blocking send with a `select`:

```go
select {
case done <- true:
default:
}
```

This way, if nobody is listening or the channel is already full, the send doesn’t block. Without this, the handler never returns, so the routine that triggered it stalls forever.

This simple change fixes the problem and allows triggering the termination event outside PocketBase as intended.

I also made the `done` channel buffered because, even though it shouldn’t happen, the write can theoretically occur before the listener starts. An unbuffered channel would skip the send if no listener existed yet; if the listener starts afterwards, the signal would be lost.

---

Thank you very much for this great project. I really love it! :heart:

### DIFF
--- a/core/base.go
+++ b/core/base.go
@@ -1408,7 +1408,7 @@ func getLoggerMinLevel(app App) slog.Level {
 func (app *BaseApp) initLogger() error {
 	duration := 3 * time.Second
 	ticker := time.NewTicker(duration)
-	done := make(chan bool)
+	done := make(chan bool, 1)
 
 	handler := logger.NewBatchHandler(logger.BatchOptions{
 		Level:     getLoggerMinLevel(app),
@@ -1479,7 +1479,10 @@ func (app *BaseApp) initLogger() error {
 
 			ticker.Stop()
 
-			done <- true
+			select {
+			case done <- true:
+			default:
+			}
 
 			return e.Next()
 		},


ganigeorgiev:
I'll have a more detailed look at the report later after work but to just make sure that I understand the problem - you are experiencing a deadlock when manually triggering `app.OnTerminate().Trigger(&core.TerminateEvent{App: app})`, is that correct?

In that case, could you create a minimal reproducible?

yerTools:
Yes, that is the problem.

Here is a small example you can run with `go run main.go serve --http=0.0.0.0:8161`:

```go
// go run main.go serve --http=0.0.0.0:8161

package main

import (
	"sync"
	"time"

)

func main() {
	app := pocketbase.NewWithConfig(pocketbase.Config{
		DefaultDev: false,
	})

	wg := sync.WaitGroup{}
	wg.Add(1)

	go func() {
		defer wg.Done()

		println("Starting PocketBase ...")
		app.Start()

		panic("unreachable")
	}()

	println("Waiting for 3 seconds ...")
	<-time.After(3 * time.Second)

	println("Stopping PocketBase ...")
	app.OnTerminate().Trigger(&core.TerminateEvent{App: app})
	println("Stop event sent")

	wg.Wait()

	panic("unreachable")
}

// Waiting for 3 seconds ...
// Starting PocketBase ...
// 2025/11/07 03:16:07 Server started at http://0.0.0.0:8161
// ├─ REST API:  http://0.0.0.0:8161/api/
// └─ Dashboard: http://0.0.0.0:8161/_/
// Stopping PocketBase ...
// Stop event sent
// {Deadlock here, no panic occurs.}
```

The output is the following (ignoring some PocketBase logging):

> Waiting for 3 seconds ...  
> Starting PocketBase ...  
> 2025/11/07 03:16:07 Server started at http://0.0.0.0:8161  
> ├─ REST API:  http://0.0.0.0:8161/api/  
> └─ Dashboard: http://0.0.0.0:8161/_/  
> Stopping PocketBase ...  
> Stop event sent

After that the app stalls "forever" and there isn't any panic. It doesn't do anything and a <kbd>Ctrl</kbd> + <kbd>C</kbd> also doesn't do anything.  
`app.Start()` never returns.

ganigeorgiev:
Thank you for the reproducible.

When I first read the PR I though you meant the `done` channel from the first code snippet and I was confused what the logger initialization has to do with that _(there is really no need to copy chunks of the source code or the fix in the PR description as it just adds unnecessery noise; one or two sentences about the fix and a reproducible are usually more than enough)_.

As you've found out the issue is caused when `OnTerminate` is triggered multiple times because after the first call there will be no one left reading from the logger's `done` channel.

The suggested fix to make the send non-blocking seems correct to me. I've merged it locally in the `develop` branch and added a small test. The fix will be part of the next release sometime later this weekend.

---

_Side-note1: Eventually we could also consider skipping the internal `OnTerminate` trigger if `OnTerminate` was already manually invoked and just call the finalizer to avoid duplicating the attached hook listeners, but I'm not sure whether it is worth the extra complexity around the state tracking and will have to think a little more on it._

_Side-note2: For internal/localhost only development, I would suggest to avoid binding to the wildcard `0.0.0.0` and instead prefer `127.0.0.1` because otherwise you may end up accidentally exposing your localhost (unless this is what you want to do). For the reason about the confusing default flag values, see https://github.com/pocketbase/pocketbase/discussions/3190._

Issue #7329, opened 2025-11-16
Bug: Rate Limiter never resets


svobol13:
Hello I am using rate limiter beta feature. When polling (every 3 seconds) burstable endpoint (something like GET /entity/status) I want configuration like 30 seconds and 30 requets. Current implementation of rate limiter doesn't support such a usecase because in that case the available budget never resets.

ganigeorgiev:
I don't understand what is the issue and why do you consider the current behavior a bug.

Could you elaborate more and provide a minimal reproducible?

ganigeorgiev:
I'm sorry but I don't see how this is a bug.

There are multiple ways to implement a rate limiter and while you may disagree with how it works in PocketBase, that doesn't make it a bug (or at least I don't understand what is the issue here).

If there is enough demand and the use cases seems reasonable and general enough we can eventually consider changing the algorithm and/or add an option for the users to chose from multiple implementations but as of now I don't see a good enough reason to change it _(if you disagree then elaborate more clearly on your use case)_.

svobol13:
@ganigeorgiev Minimal reproducible case is:
- Define endpoint with window 10s and number of requests 5.
- Define a client, that will call that endpoint every 8 seconds.
- Expected behaviour: Since this cannot ever exceed 2 requests within any 10s interval, it should never return 429
- Actual behaviour: Since it never resets the time window, it will return 429 every 6th request and it will reset after 10 seconds of 429s.

Current behaviour of the limiter is weird, because "duration" means "never reset counter, unless there is a pause between requests that is longer than duration configured".

Current implementation claim it is "fixed window" RL. But it is not because beginning of current window moves with every request - it is not fixed window. Current implementation could be called "stretching fixed window" - I have never heard of such an algorithm and cannot figure out a use case for it (you need to grant users huge limits, to prevent false-positives in case of "quickly clicking" in the UI and thus the RL is rather useless, or you need to rely on users doing "pauses" long enough to reset the budget). That is why I think this is rather a bug and not an alternative implementation.

ganigeorgiev:
Thank you for providing more details. I see from where the confusion comes from _(don't stick to the internal struct name as the limiter implementation has changed several times since its initial release)_.

I've added it in my todo to eventually reconsider changing the tracking together with the option to exclude IPs but for now this remains a very low priority.

svobol13:
[...]

PR #7330, opened 2025-11-16
Modified rate limiter reseting policy. See #7329

svobol13:
Fixes "burstable polling" issue with ratelimiter: https://github.com/pocketbase/pocketbase/issues/7329


ganigeorgiev:
I don't think this is needed.

_Edit: the user provided more details about the change in the related issue. I'll consider reavulate it in the near future as it is probably more easy to reason and closer to the user expectations based on how the configuration is shown in the UI._

PR #7335, opened 2025-11-19
Add excluded IPs option for rate limiting

muhamd11:
This adds the ability to exclude specific IP addresses from rate limiting.

I noticed there wasn't a way to whitelist certain IPs (like internal services or monitoring tools) from the rate limit rules, so I added an `excludedIPs` field to the rate limit settings.

Changes:
- Added `ExcludedIPs` field to the rate limit settings model
- Updated the rate limit middleware to check if an IP is in the excluded list before applying limits
- Added a UI field in the rate limit accordion to manage excluded IPs

The excluded IPs are checked before any rate limit rules are applied, so whitelisted addresses won't count against the limits at all.


ganigeorgiev:
Thank you but once again PRs for new features are not welcomed and I'll appreciate if users read the contibuting note in the readme and more specifically:

> But please refrain creating PRs for new features without previously discussing the implementation details. PocketBase has a [roadmap](https://github.com/orgs/pocketbase/projects/2) and I try to work on issues in specific order and such PRs often come in out of nowhere and skew all initial planning with tedious back-and-forth communication.

This feature request already exists for later evaluation but for now remains a very low priority. If you want it to get prioritized then a better approach would be to explain in more details what actually is your use case, aka. what are these "internal services or monitoring tools" and how and why exactly they would interact with PocketBase (if it is as superuser then the rate limiter is already ignored).

PR #7353, opened 2025-11-25
Fix: autocomplete for modifiers breaks down at the ":" char

ouvreboite:
When writing rules/filters, the autocomplete display correctly the possible modifier variants. But when the ":" character is typed, the autocomplete stops. It resumes only when two characters of the possible modifier(s) are typed.

Image

I think that's due to the fact ":" is not part of the word regex in FilterAutocompleteInput.svelte, so the modifier is considered as a new word.
   https://github.com/pocketbase/pocketbase/blob/90d896e1cc49e3535dad95718dada3f7efcef4ca/ui/src/components/base/FilterAutocompleteInput.svelte#L247

I updated the regex locally to `let word = context.matchBefore(/[\'\"\@\w\.:]*/);` and it seems to be working fine ⬇️ 

Image

I'm not seeing UI tests, so I'm worried that the fix itself could have some side effects.
I can transform this PR into an issue if you prefer to take the time to look into that yourself.

### DIFF
--- a/ui/src/components/base/FilterAutocompleteInput.svelte
+++ b/ui/src/components/base/FilterAutocompleteInput.svelte
@@ -244,7 +244,7 @@
 
     // Returns object with all the completions matching the context.
     function completions(context) {
-        let word = context.matchBefore(/[\'\"\@\w\.]*/);
+        let word = context.matchBefore(/[\'\"\@\w\.:]*/);
         if (word && word.from == word.to && !context.explicit) {
             return null;
         }


ganigeorgiev:
Thank you. I'll consider it eventually merging it locally as the ui/dist also needs to be regenerated with the next minor release but for now I don't think there is need for this (note that also the UI is in process in being rewritten to allow external customizations; there are no ETAs).

PR #7372, opened 2025-12-05
Fix S3 0-byte download for Gzip files

hitomi-krgr:
This PR fixes an issue where downloading Gzip-compressed files from S3 storage would result in a 0-byte file.

### The Issue
When using the S3 driver, Go's `http.Client` (used internally by the AWS SDK) transparently handles Gzip compression. If the S3 response contains `Content-Encoding: gzip`, the client automatically decompresses the body.

As a side effect of this transparent decompression, the `Content-Length` header is removed from the response because the uncompressed size is unknown during the stream. Consequently, `blob.Reader.Size()` returns `0` (or `-1`).

This behavior is documented in the [Go `net/http` Transport documentation](https://pkg.go.dev/net/http#Transport) (under `DisableCompression`):
> "If the Transport requests gzip on its own and gets a gzipped response, it's transparently decoded in the Response.Body. However, if the user explicitly requested gzip it is not automatically uncompressed."

When `http.ServeContent` receives a reader with size `0`, it serves an empty response, resulting in a 0-byte download for the user.

### The Fix
This change adds a generic fallback mechanism in `filesystem.Serve`:

1.  It checks if `br.Size() <= 0`.
2.  If the size is unknown (which happens during transparent decompression), it bypasses `http.ServeContent` and uses `io.Copy` to stream the content directly to the response.
3.  It manually sets the `Last-Modified` header if available, to preserve basic metadata.

This ensures that Gzip files stored in S3 are served correctly while maintaining standard behavior## Fix S3 0-byte download for Gzip files

This PR fixes an issue where downloading Gzip-compressed files from S3 storage would result in a 0-byte file.

### The Issue
When using the S3 driver, Go's `http.Client` (used internally by the AWS SDK) transparently handles Gzip compression. If the S3 response contains `Content-Encoding: gzip`, the client automatically decompresses the body.

As a side effect of this transparent decompression, the `Content-Length` header is removed from the response because the uncompressed size is unknown during the stream. Consequently, `blob.Reader.Size()` returns `0` (or `-1`).

This behavior is documented in the [Go `net/http` Transport documentation](https://pkg.go.dev/net/http#Transport) (under `DisableCompression`):
> "If the Transport requests gzip on its own and gets a gzipped response, it's transparently decoded in the Response.Body. However, if the user explicitly requested gzip it is not automatically uncompressed."

When `http.ServeContent` receives a reader with size `0`, it serves an empty response, resulting in a 0-byte download for the user.

### The Fix
This change adds a generic fallback mechanism in `filesystem.Serve`:

1.  It checks if `br.Size() <= 0`.
2.  If the size is unknown (which happens during transparent decompression), it bypasses `http.ServeContent` and uses `io.Copy` to stream the content directly to the response.
3.  It manually sets the `Last-Modified` header if available, to preserve basic metadata.

### DIFF
--- a/tools/filesystem/filesystem.go
+++ b/tools/filesystem/filesystem.go
@@ -462,6 +462,15 @@ func (s *System) Serve(res http.ResponseWriter, req *http.Request, fileKey strin
 	// that are made in the last day while revalidating the res in the background)
 	setHeaderIfMissing(res, "Cache-Control", "max-age=2592000, stale-while-revalidate=86400")
 
+	// fallback to io.Copy if the size is unknown/empty (eg. due to transparent gzip decompression)
+	if br.Size() <= 0 {
+		if !br.ModTime().IsZero() {
+			res.Header().Set("Last-Modified", br.ModTime().UTC().Format(http.TimeFormat))
+		}
+		_, err := io.Copy(res, br)
+		return err
+	}
+
 	http.ServeContent(res, req, name, br.ModTime(), br)
 
 	return nil


hitomi-krgr:
I initially considered solving this by disabling transparent decompression in the S3 client (DisableCompression: true). This would allow us to retrieve the correct Content-Length (of the compressed file) and stream the raw Gzip data, which would work nicely with http.ServeContent.

However, since the filesystem package currently handles tasks like CreateThumb that expect uncompressed image data, disabling decompression at the transport layer might interfere with those operations.

While separating image processing from the storage layer could be a long-term goal, for now, the io.Copy fallback seems like a safer, minimally invasive fix that resolves the download issue without affecting other parts of the system.

hitomi-krgr:
### Reproduction Steps (Using MinIO)

This issue can be reproduced locally using MinIO to simulate an S3 environment serving files with `Content-Encoding: gzip`.

#### 1. Setup Environment
Start a local MinIO instance and configure PocketBase to use it as S3 storage.
- **MinIO Endpoint**: `http://localhost:9000`
- **Bucket**: `test-bucket`

#### 2. Upload File via PocketBase
Log in to the PocketBase Admin UI and upload a file (e.g., `image.png`) to any collection record normally.
- Note the **Collection ID**, **Record ID**, and **Filename**.

#### 3. Inject "Problematic" Metadata
Use `mc` (MinIO Client) to modify the file's metadata in the bucket, simulating an external process (like a CDN or Lambda) that compressed the file.

```bash
# Configure alias
mc alias set local http://localhost:9000 minioadmin minioadmin

# Path format: bucket/collectionId/recordId/filename
# Example: test-bucket/pbc_1234567890/r_1234567890/image.png

# Overwrite the file with itself, adding the Content-Encoding: gzip header
# (Note: In a real scenario the content would be gzipped, but setting the header 
# is enough to trigger Go's transparent decompression logic which strips Content-Length)
mc cp local/test-bucket/COLLECTION_ID/RECORD_ID/FILENAME \
      local/test-bucket/COLLECTION_ID/RECORD_ID/FILENAME \
      --attr "Content-Encoding=gzip"
```

#### 4. Trigger the Bug
Try to download the file via PocketBase API.

```bash
curl -v "http://localhost:8090/api/files/COLLECTION_ID/RECORD_ID/FILENAME"
```

**Before Fix (Bug):**
The response returns `200 OK` but with `Content-Length: 0`. The download is empty because `http.ServeContent` cannot handle the stream when `Content-Length` is stripped by Go's http client (due to the gzip header).

**After Fix:**
The file content is served correctly using `io.Copy` fallback.

ganigeorgiev:
There is way too many text for an issue that I'm not even quite sure that it needs fixing in PocketBase. I'll investigate it later this weekend in more details but this AI generated PRs are really not welcomed. It would have been better to just explain what is actually the problem and when it occur instead of generating this wall of text that I would have to verify.

hitomi-krgr:
Sorry, I'll close this and submit a cleaner PR

ganigeorgiev:
@hitomi-krgr Please don't use AI. Your help is not needed. I'll investigate and verify the issue on my own (it needs tests anyway).

PR #7382, opened 2025-12-09
Add horizontal scaling plugin

mattzab:
This is my first time opening a pull request. I read your contributing guidelines and I think this follows them.
I'm sure there's a cleaner way to do a pull request, I just don't know what it is, unfortunately. I'm just one guy tinkering around. I mean not to waste your time. Feel free to reject this if it is not within scope or in any other way unusable.

This is a plugin for pocketbase to add horizontal scaling via NATS Jetstream. It is leaderless, eventually consistent, event-driven, and syncs both records and schema.

Event-driven is a benefit over other resource-heavy solutions like polling the WAL journal.
Leaderless is a nice perk over some of the overlayFS style solutions where there's essentially a central node that all other nodes follow.

Realtime websockets (unfortunately untested at this time) are handled by websocket routing. Basically, if an end-user opens a realtime subscription, this is published to all other instances. Any other user that wants to open a realtime sub to the same resource will have their websocket routed to the instance that already has an open sub.

Feel free to give me your candid response. I'm new, and I wanted to build this as a proof-of-concept for something that I think would be really cool within Pocketbase, if that's feasible.

Thanks for everything you do, Pocketbase is awesome and has replaced all of my manual sqlite operations.
Regards,
Matt


ganigeorgiev:
I haven't checked the code but I'm not interested in this feature.

There is nothing wrong of maintaining your own fork or moving the plugin part in its own standalone Go package _(after the ongoing UI rewrite there will be also an option for external UI elements for full fledged frontend and backend plugin functionality; there are no ETAs)_.

PR #7386, opened 2025-12-11
dockerized the project for ease of development and deployment

onixldlc:
now... I'm not sure if throwing dockerization into the mix is a good idea, since the project are currently still in its infancy, but  since backward compatibility is not guaranteed, i think this may help other to see if changing to the latest version would have minor or significant changes to their setup, atleast for static or js extend that is.

although with that being said... i  mainly want to use this as a quick and dirty way to quickly build out a prototype without cluttering my go pkg folder :grin:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.dockerignore
+docker-compose.yml
+Dockerfile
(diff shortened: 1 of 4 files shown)


ganigeorgiev:
This type of PRs are not welcomed here.

_And please, next time don't add Copilot or other AI agent as a reviewer because you'll get banned._

PR #7392, opened 2025-12-16
perf(security): upgrade to math/rand/v2

rhzs:
The old `math/rand` uses a global lock on every call. Under concurrent load (which is pretty much every request since `PseudorandomString` is used for filter placeholders, hook IDs, query aliases, etc.), this becomes a bottleneck.

`math/rand/v2` uses per-goroutine state so there's no lock contention.

See: [Go 1.22 rand/v2 blog post](https://go.dev/blog/randv2) and [math/rand/v2 docs](https://pkg.go.dev/math/rand/v2)

### DIFF
--- a/tools/security/random.go
+++ b/tools/security/random.go
@@ -3,7 +3,7 @@ package security
 import (
 	cryptoRand "crypto/rand"
 	"math/big"
-	mathRand "math/rand" // @todo replace with rand/v2?
+	mathRand "math/rand/v2"
 )
 
 const defaultRandomAlphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
@@ -52,7 +52,7 @@ func PseudorandomStringWithAlphabet(length int, alphabet string) string {
 	max := len(alphabet)
 
 	for i := range b {
-		b[i] = alphabet[mathRand.Intn(max)]
+		b[i] = alphabet[mathRand.IntN(max)]
 	}
 
 	return string(b)


ganigeorgiev:
The todo exists for a reason. It will be evaluated at some point in the future, but I have to read more properly over the changes and for now remains a low priority (see also https://github.com/pocketbase/pocketbase/pull/6045, https://github.com/pocketbase/pocketbase/pull/7294; _I'll consider prioritizing it with the next minor release as people seem to spam me about this_).

_Side-note since you've opened another performance related PR - it will be helpful to provide some real numbers to evaluate better performance motivated changes and their trade offs. For example in this particular case, your claim that this is a bottleneck may not exactly be accurate because Go 1.20+ has backported some of the v2 improvements in the v1 code of the package. **If this and your other PRs are a result of AI suggestion, then please don't do that again because I don't want to collaborate with people like that.**_

rhzs:
I'm more than welcome to provide with the benchmark numbers and profiling data—that's a fair ask for the PR. However, I'd like to address the other part of your comment. 
Stating that "**don't want to collaborate with people like that**" based on an assumption about how code was written is both **presumptuous** and **unwelcoming**. Developers use IDEs, autocomplete, Stack Overflow, documentation, and yes—AI assistants. _Do you code using pen & paper no help at all?_

The bar for contribution should be code quality and the contributor's ability to engage meaningfully, not _**policing**_ their workflow. Whether we use AI tooling or not is irrelevant—what matters is whether we understand the code we are submitting, can defend it technically, and can iterate on feedback. This what matters most. Feel free to differ.

ganigeorgiev:
> Whether we use AI tooling or not is irrelevant

No. If you think this way then please use something else but this contribution is NOT helpful in any way (especially as there are already 2 other PRs for it).

**And there is already a `todo` addressing it.** I can replace the import on my own if I wanted to. It needs to be carefully evaluated because again upgrading just for the sake of the upgrade without actually understanding what has changed is wrong (did you measure whether the new chacha default algorithm actually performs better than the pcg? as it is currently is it really an actual bottleneck?).

In any case, I've already added a note to evaluate the upgrade but I'm not going to waste further time here. Maintain your own fork and stop spamming me.

PR #7393, opened 2025-12-16
perf(core): optimize FieldsList.MarshalJSON

rhzs:
Use `bytes.Buffer` and direct JSON splicing instead of unmarshaling to `map[string]any`, reducing allocations in JSON serialization.

### DIFF
--- a/core/fields_list.go
+++ b/core/fields_list.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"bytes"
 	"database/sql/driver"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
No sorry. This actually was considered in the past but the current implementation is less error prone and I prefer to stay as it is unless the benefit is significant to make a noticeble difference. Note that in this particular case this is used primarily for the Collections APIs and the performance here is not that important (the related Record APIs uses a cached state of the collection).

PR #7462, opened 2026-01-18
feat: Native soft delete support

### LINKED ISSUE #2866: Feature Request: Implement soft delete by default
# Use Case:
Soft delete is a common feature that allows data to be marked as deleted without actually removing it from the database. This enables the recovery of mistakenly deleted data, and also maintains the historical record for audit purposes. Currently, users have to manually implement this feature, which can lead to inconsistencies and extra work.

# Description:
We propose to implement soft delete by default for every collection. Users should also have the ability to deactivate this feature if desired. Currently, tools like Gorm (insert link to the documentation here) use a "deleted" timestamp and filter out all issues where this timestamp is set. We suggest a similar approach.

# Implementation:
The main aspect of the implementation will be adding a "deleted" timestamp column to all collections. This column will mark when an item was "deleted", without actually removing the item from the database. Secondly, all API functions that retrieve data will need to be updated to filter out the items marked as deleted, unless specifically requested.

# Requirements:
- Add a "deleted" timestamp column to all collections.
- Modify all API retrieval functions to filter out items marked as deleted.
- Provide an option for users to deactivate the soft delete feature.
- Update the API to include functionality that allows retrieval of items marked as deleted when specifically requested.

# Advantages:
- Consistent functionality across all collections.
- Users can recover mistakenly deleted data.
- Decreased need for manual implementation by users.
- Maintains a historical record of all data, which can be beneficial for auditing purposes.

# Workaround for users that will implement it on there own:
Currently, users can manually implement this feature by adding the "deleted" timestamp column and modifying their queries to filter out items where this timestamp is set. However, this can lead to inconsistencies and adds an extra layer of work for the users. The proposed implementation would streamline the process and ensure consistent functionality across all collections.

joachimBrindeau:
## Summary

This PR implements native soft delete support for PocketBase collections, addressing issue #2866.

**Key Features:**
- Collection-level `softDelete` boolean toggle
- When enabled, DELETE operations set `deleted` timestamp instead of removing records
- Queries automatically filter out soft-deleted records by default
- API `showDeleted=true` parameter (superuser-only) to include deleted records in list responses

## Design Philosophy

**Minimal & Unopinionated:**
- No cascade soft delete - relations work as before
- No automatic restore API - records can be restored via normal UPDATE
- No special "trash" views - use `showDeleted=true` if needed
- Same API endpoints, same responses - behavior change is internal only

### Core (`core/`)
- `collection_model.go`: Added `SoftDelete bool` field to collection model
- `field.go`: Added `FieldNameDeleted = "deleted"` constant
- `record_model.go`: Modified `onRecordDeleteExecute` to UPDATE instead of DELETE when SoftDelete enabled
- `record_query.go`: Added filter to exclude soft-deleted records; added `RecordQueryIncludeDeleted()` method
- `app.go`: Added `RecordQueryIncludeDeleted` to App interface

### API (`apis/`)
- `record_crud.go`: Added `showDeleted` query parameter support (superuser-only)

### Migrations (`migrations/`)
- `1768756423_add_soft_delete.go`: Migration for existing installations
- `1640988000_init.go`: Updated for new installations

## Usage

```go
// Enable soft delete on a collection
collection := core.NewBaseCollection("posts")
collection.SoftDelete = true
app.Save(collection)

// DELETE now sets `deleted` timestamp instead of removing
app.Delete(record) // Record still exists with deleted timestamp

// Normal queries exclude deleted records
records, _ := app.FindAllRecords("posts") // Only non-deleted

// Superusers can see deleted via API
// GET /api/collections/posts/records?showDeleted=true
```

## Test Plan

- [x] Unit tests for `SoftDelete` collection field
- [x] Unit tests for auto-initialization of `deleted` field
- [x] Unit tests for soft delete behavior (UPDATE instead of DELETE)
- [x] Unit tests for query filtering
- [x] Unit tests for `RecordQueryIncludeDeleted`
- [x] API tests for `showDeleted` parameter
- [x] All existing tests pass

## Related Issue

Closes #2866

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ganigeorgiev:
No. Please read the contribution note in the README.

_And if you open again AI generated PRs you'll get banned because I don't want to collaborate with people like that._

PR #7457, opened 2026-01-17
fix: typos in comments

NAM-MAN:
## Summary

Fix typos in comments:

- `tools/filesystem/internal/s3blob/s3blob.go`: `the the internal` → `the internal`
- `core/settings_model.go`: `the the max` → `the max` (2 occurrences)

PR #7469, opened 2026-01-23
chore: Update outdated GitHub Actions versions

pgoslatara:
This PR updates outdated GitHub Action versions.

- Updated `actions/setup-node` from `v4` to `v6` in `.github/workflows/release.yaml`
- Updated `actions/checkout` from `v4` to `v6` in `.github/workflows/release.yaml`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,12 +16,12 @@ jobs:
         run: echo "flags=--snapshot" >> $GITHUB_ENV
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
 
       - name: Set up Node.js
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Thank you but there is no need for this at the moment. If you think otherwise elaborate what is the issue with the current used versions _(if it is regarding the node action deprecation planned for this year, then it will be addressed later)_.

Note that PocketBase is currently in the process of a major refactoring and the actions setup will change anyway in the near future.

As a side-note, similar to https://github.com/pocketbase/pocketbase/pull/7457, such low effort AI PRs are not welcome here so please don't do that again or you'll get banned _(you've opened at least 100 PRs to many other repos that I'm watcher and it is not too much different from a spam campaign; if somebody wants to use dependabot they can just enable it, your help is not needed)._

PR #7484, opened 2026-01-28
Fixed data races found on concurrency tests

wongak:
Added mutexes for accessing expectations in tests.

### DIFF
--- a/tools/cron/cron_test.go
+++ b/tools/cron/cron_test.go
@@ -3,6 +3,7 @@ package cron
 import (
 	"encoding/json"
 	"slices"
+	"sync"
 	"testing"
 	"time"
 )
@@ -253,6 +254,7 @@ func TestCronJobs(t *testing.T) {
 func TestCronStartStop(t *testing.T) {
 	t.Parallel()
 
+	var m sync.Mutex
 	test1 := 0
 	test2 := 0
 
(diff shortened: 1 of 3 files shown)


ganigeorgiev:
Thank you but I prefer to investigate it in more details and eventually will consider applying them manually in the `develop` branch because some of the suggestions seems unnecessery or require larger tests refactoring.

In any case instead of just "merge this", it will help to elaborate what are the issues the current code is causing and what was the reason the PR was opened in the first place.
Is it test flakiness? Tests run with `-race`? For both cases adding mutexes unfortunately will not resolve it fully because there are other "flawed" tests that will not be influenced by this change.

_If this was suggested by AI, then please don't open PRs like that again._

wongak:
Sorry, forgot to add more info.

Yes, these fix the issues encountered when running tests with `-race`.

The race detector might have false negatives, but no false positives. So the encountered races are definitely issues.

But I don't think other tests will be influenced. The issues are related to accessing variables set up in tests, which are subsequently mutated in goroutines. The mutexes are only guarding the test setup.

_And no, I am not using AI. I detest those._

tools/cron/cron_test.go
```diff
@@ -280,12 +286,14 @@ func TestCronStartStop(t *testing.T) {
 	c.Stop()
 	c.Stop()
 
+	m.Lock()
 	if test1 != expectedCalls {
```
wongak: accessed here

ganigeorgiev:
> But I don't think other tests will be influenced. 

It is not about other tests being influenced. It is that there are other tests that will fail when run with `-race` (e.g. some API related ones that has fireAndForget goroutine calls; there are plans to replace them in the near future).

I'll review and apply the changes manually in the `develop` branch but again there is no guarantee that all tests will pass with `-race`.

PR #7510, opened 2026-02-08
feat: support constant array operands in API rules

### LINKED ISSUE #6308: Feature Request: constant array in API rules
I don't really know if it's a feature request or it is already possible and I'm missing some sintax.

I want to set a rule to allow authenticated users with the role "isSecretary" or "isAccountant".
the role is stored as a multi-select field "role" in the auth collection.
So i should write:

`@request.auth.id !='' &&( @request.auth.role ?= “isSecretary” ||  @request.auth.role ?= “isAccountant”)`

this is just an example but there are more || roles to check. 
I was wondering if there is a better way to do it and shorten the rule like:

`@request.auth.id !='' &&  @request.auth.role:each ?= [ “isSecretary”, ”isAccountant” ]`

apk471:
This PR adds support for constant array operands in PocketBase API rules.

It allows concise checks against multiple allowed values without repeating
OR conditions, for example:

@request.auth.role ?= array("isSecretary", "isAccountant")

Existing scalar `?=` behavior remains unchanged and backward compatible.
The array variant evaluates to true when there is at least one common value
between the field and the constant array.

Tests have been added to cover parsing and evaluation behavior.

Closes #6308

### DIFF
--- a/core/record_field_resolver_runner.go
+++ b/core/record_field_resolver_runner.go
@@ -777,7 +777,8 @@ func (r *runner) finalizeActivePropsProcessing(collection *Collection, prop stri
 	// -------------------------------------------------------
 	if modifier == eachModifier && isMultivaluer {
 		jePair := r.activeTableAlias + "." + cleanFieldName
-		jeAlias := "__je_" + r.activeTableAlias + "_" + cleanFieldName + r.resolver.joinAliasSuffix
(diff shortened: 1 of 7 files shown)


ganigeorgiev:
I'm sorry but this is not welcome. Please use your own fork and don't spam further.

PR #7525, opened 2026-02-14
Fix case-sensitive bearer token prefix stripping

benjamesfleming:
The RequireAuth middleware is handing the `"Bearer "` prefix in the authorization header in a case-senstive way. This caused issues for me when running the OpenID Connect conformance tests against the [PocketBase OAuth2 Plugin](https://github.com/benjamesfleming/pocketbase-ext-oauth2) I'm building. For the plugin, I'm hoping to keep the native PocketBase authentication stack by using normal Access Tokens + RequireAuth().

This pull request updates the `"Bearer "` prefix handling to be case-insensitive and adds tests to ensure correct handling.

Reference: https://auth0.com/blog/the-bearer-token-case/

### DIFF
--- a/apis/middlewares.go
+++ b/apis/middlewares.go
@@ -210,9 +210,20 @@ func getAuthTokenFromRequest(e *core.RequestEvent) string {
 	if token != "" {
 		// the schema prefix is not required and it is only for
 		// compatibility with the defaults of some HTTP clients
-		token = strings.TrimPrefix(token, "Bearer ")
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
Thank you! I've pushed a slightly different version in the `develop` branch and it will be part of the next minor release soon _(most likely next week)_.

_I've skipped the whitespace trimming of the header values because it is expected to be already stripped for us by the Go std request parsing but feel free to let me know if it is causing you issue and I'll add it._

PR #7532, opened 2026-02-17
fix(ime): add input state tracking to prevent composition truncation

miaopan607:
fix #7530


ganigeorgiev:
I feel that I don't understand the problem or I don't know how IME is being used.

Even if we temporary skip the normalization to allow IME entry, as mentioned previously the collections and fields name currently supports only characters in the range `[a-zA-Z0-9_]`. Or in other words - what is the problem that you actually have and how the above helps? Why IME is even needed here?

miaopan607:
I understand that the fields only support `[a-zA-Z0-9_]`. However, constantly switching the IME on and off just to type a few English words is quite inconvenient for non-native speakers. Many of us stay in IME mode because it also provides autocorrect and predictive suggestions for English.

For example,
image


The core issue is that the UI interrupts the input process prematurely. Regardless of the character limit, fixing this would ensure a smoother and more consistent typing experience for everyone. It's a UX improvement that prevents the input from being "broken" by the system.

ganigeorgiev:
> Many of us stay in IME mode because it also provides autocorrect and predictive suggestions for English.

OK, that makes sense.

I've pushed a slightly different version in the `develop` branch that just check for the `e.isComposing` field of the `input` event (_it will be part of the next minor release, most likely later this weekend_).

Please test it when you find time and let me know if it doesn't work for you.

ganigeorgiev:
Oh, relying only on `e.isComposing` in this case will not work because if the user type the suggestion fully and select it it will not apply the input normalization because there will be no new input. So we'll need also the `compositionend` event (we may also have to check the composition data in case the composition was discarded/cancelled).

I'll push an update in the `develop` branch ~a little bit later~ _Done_.

miaopan607:
I’ve tested it on my side and the issue seems to be resolved now. Thank you!

PR #7534, opened 2026-02-19
Remove max-width constraint from backup-name

ajnart:
After 300px the backup names start to displays an ellipsis (...) , I'm just removing this from the CSS so that we can see the full name of the backup in screens that are wide enough 

 

<table>
<tr>
 <td> Before
 <td> After
<tr>
 <td> 
bd30c663b10fddd7a8131b502aa53f91c8b946982afe88ab9f17cc074974f37f
 <td> 
CleanShot 2026-02-19 at 10 07 35@2x

</table>

### DIFF
--- a/ui/src/components/settings/BackupsList.svelte
+++ b/ui/src/components/settings/BackupsList.svelte
@@ -214,7 +214,6 @@
         }
     }
     .backup-name {
-        max-width: 300px;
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
No, but I'll consider it expanding to the end of the row in the new UI _(Done)_.

PR #7555, opened 2026-03-02
Simplify code

deining:
This PR simplifies the code by using `max` function.


ganigeorgiev:
Thank you. I'll consider applying this and other minor new Go features in the near future but there is no need for that at the moment.

_Edit: I've received several notifications for similar AI assisted changes in other repos I watch and similar to PRs like in https://github.com/pocketbase/pocketbase/pull/7469, I'm just going to block people as I don't want to encourage it anymore._

PR #7575, opened 2026-03-08
Fix high memory usage in RereadableReadCloser

s1adem4n:
This fixes the high memory usage of the RereadableReadCloser as discussed in #7572

The reader now switches to a temporary file after 10mb were read (similar to how the [multipart package](https://cs.opensource.google/go/go/+/refs/tags/go1.26.1:src/mime/multipart/formdata.go;l=41) does it)

When uploading a 2gb file with the old version the memory usage spiked to ~3gb for me, the new version stays stable at ~50mb :)
Here are the pprof heap details if you want to double-check:
[new_profile.pb.gz](https://github.com/user-attachments/files/25824937/new_profile.pb.gz)
[old_profile.pb.gz](https://github.com/user-attachments/files/25824938/old_profile.pb.gz)

I had to add a deferred `Close()` to all the places where the reader was used so the temporary file gets deleted. 
What you may need to confirm is whether to also close the underlying reader in the same `Close()` function.

Lastly I added a new test that confirms the logic and ensures the temporary file is also deleted when it is created.

### DIFF
--- a/apis/batch.go
+++ b/apis/batch.go
@@ -197,7 +197,6 @@ func (p *batchProcessor) Process(batch []*core.InternalRequest, timeout time.Dur
 
 		go func() {
 			err := p.process(txApp, batch, 0)
-
 			if err != nil {
 				err = validation.Errors{
 					"requests": validation.Errors{
@@ -261,7 +260,6 @@ func (p *batchProcessor) process(activeApp core.App, batch []*core.InternalReque
 				return err
 			},
 		)
-
 		if err != nil {
 			return err
 		}
@@ -364,6 +362,7 @@ func processInternalRequest(
 	// assign request
 	event.Request = r
 	event.Request.Body = &router.RereadableReadCloser{ReadCloser: r.Body} // enables multiple reads
+	defer event.Request.Body.Close()
 
 	// assign response
 	rec := httptest.NewRecorder()
--- a/tools/router/rereadable_read_closer.go
+++ b/tools/router/rereadable_read_closer.go
@@ -1,9 +1,11 @@
 package router
 
 import (
-	"bytes"
 	"io"
+	"os"
 )
+
+const defaultMaxMemory = 10 << 20 // 10 MB
 
 var (
 	_ io.ReadCloser = (*RereadableReadCloser)(nil)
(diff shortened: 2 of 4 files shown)


ganigeorgiev:
Thank you for looking at it. At first glance it looks OK but there are some minor details that I'm not sure about at the moment:

- the extra offset read/write tracking and flags makes it a little bit difficult to follow and if possible I would prefer to avoid it; eventually we can also move it into its own `BufferWithFile` struct that would implement the `io.ReadWriteCloser` interface
- if possible avoid copying back the 10mb memory buffer into the file and instead treat the file more like an "extension" (aka. first write in memory, then in file; for reads it could be identical - first read from memory and then continue from the file; eventually also leverage the std `io.MultiReader`/`io.MultiWriter`)
- ideally we should collect all errors (including those from reread) and return them together with the final close

Or in other words, a very rough outline/pseudo-code of what I had in mind could look something like:

```go
type RereadableReadCloser struct {
    io.ReadCloser

    copy       io.ReadWriteCloser
    closeFuncs []func() error
}

func (r *RereadableReadCloser) Read(b []byte) (int, error) {
    if r.copy == nil {
        r.copy = newBufferWithFile(defaultMaxMemory)
    }

    n, readErr := r.ReadCloser.Read(b)
    if n > 0 {
        var writeErr error
        n, writeErr = r.copy.Write(b[:n])
        if writeErr != nil {
            return n, writeErr
        }
    }

    if errors.Is(readErr, io.EOF) {
        r.Reread()
    }

    return n, readErr
}


func (r *RereadableReadCloser) Reread() {
    if r.copy == nil || r.copy.Len() == 0 {
        return
    }

    // close later, together with everything else
    r.closeFuncs = append(r.closeFuncs, r.ReadCloser.Close)

    r.ReadCloser = r.copy
    r.copy = newBufferWithFile(defaultMaxMemory)
}

func (r *RereadableReadCloser) Close() error {
    var errs = make([]error, 0, len(r.closeFuncs)+2)

    if r.copy != nil {
        errs = append(errs, r.copy.Close())
    }

    for _, fn := range r.closeFuncs {
        errs = append(errs, fn())
    }

    errs = append(errs, r.ReadCloser.Close())

    return errors.Join(errs...)
}

func newBufferWithFile(limit int64) io.ReadWriteCloser {
    ...
}
```

But again, I'm not sure. Maybe your approach is better. Let's keep it for now as one version and I'll try to find time next week after work to evaluate it more carefully.

ganigeorgiev:
I've reviewed it earlier in more details and decided to push a slightly different version based on the pseudo-code above to address some of my earlier concerns. It is better also to have the `bufferWithFile` as a separate struct to make it easier in the future eventually exporting it as an util helper as there are other places that may benefit of the combined buffer.

I'll push a little bit later a RC release and will deploy it also in Presentator to monitor it.

Nonetheless, thank you for the effort and the suggested alternative.

PR #7581, opened 2026-03-11
Short-circuit top-level OR branches in rule evaluation

dtkav:
When a rule has OR branches that only reference @request.* fields (resolvable from in-memory RequestInfo), evaluate them in Go before compiling to SQL. This prevents expensive LEFT JOIN cartesian products from materializing when a cheap branch already matches.

- Add StaticResolver interface for FieldResolvers that expose static data
- RecordFieldResolver implements StaticResolver via staticRequestInfo
- Recursively simplify nested groups: A && (cheap || expensive) becomes A && (expensive) or just A when cheap matches


dtkav:
My apologies, I intended to create  PR against my own main branch.

PR #7603, opened 2026-03-30
fix Discord OAuth new username format

HansHans135:
> Reference: https://support.discord.com/hc/en-us/articles/12620128861463-New-Usernames-Display-Names

- Global name -> name
- Username -> username

### DIFF
--- a/tools/auth/discord.go
+++ b/tools/auth/discord.go
@@ -55,6 +55,7 @@ func (p *Discord) FetchAuthUser(token *oauth2.Token) (*AuthUser, error) {
 	extracted := struct {
 		Id            string `json:"id"`
 		Username      string `json:"username"`
+		GlobalName    string `json:"global_name"`
 		Discriminator string `json:"discriminator"`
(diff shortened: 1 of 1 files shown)


ganigeorgiev:
Thank you, I'll review the suggested changes and will read the article sometime later today.

ganigeorgiev:
I was able to review it before work and as far as I understand this essentially affect only the default assigned display name because we were already using just the `username` for the `Auth.Username` field _(this is because in pre v0.23 versions the username regex was fixed and the `#` character was not allowed; in case there is a duplicate we just don't assign it and generated a random one)_.

I've pushed in the `develop` branch a slightly modified version that apply the username+discriminator fallback inside the empty `global_name` check. The fix will be part of the next minor release _(sometime later ~this week/weekend~ together with the [Go 1.26.2 security release](https://groups.google.com/g/golang-announce/c/2UyQAnUQKwc/m/F3cQvlcTAgAJ))_.

HansHans135:
Thank you for the merge; I really like this project.

PR #7608, opened 2026-04-01
chore: clarify backblaze as a provider

ljesus:
- Add a Provider dropdown preset to the S3 configuration UI with support for AWS S3 and Backblaze B2
- While Backblaze B2 was already usable via the S3-compatible API, it wasn't obvious to users how to configure it, the preset auto-fills the endpoint format, enables force-path-style, and adjusts labels to B2 terminology


ganigeorgiev:
No. If you have a specific problem elaborate clearly in a Q&A discussion what is actually the issue to understand the use case but such contributions are not welcome here (especially if it was LLM generated). You are free to use your own fork, but please don't spam further.

PR #7632, opened 2026-04-09
fix: handle WalkDir err in watchHooks to prevent nil pointer dereference

mango766:
Fixes #7631

## What

The `watchHooks` function in `plugins/jsvm/jsvm.go` calls `entry.IsDir()` in a `filepath.WalkDir` callback without first checking if `entry` is nil.

Per the Go docs, when the initial `Stat` on the root directory fails, the callback receives `d == nil`. Calling `entry.IsDir()` in that case panics.

## Fix

Added an `if err != nil { return err }` guard at the top of the callback, consistent with how every other `WalkDir` callback in the codebase already handles this (see `tools/archive/create.go` and `tools/filesystem/internal/fileblob/fileblob.go`).

```diff
 dirsErr := filepath.WalkDir(watchDir, func(path string, entry fs.DirEntry, err error) error {
+    if err != nil {
+        return err
+    }
+
     // ignore hidden directories, node_modules, symlinks, sockets, etc.
     if !entry.IsDir() || entry.Name() == "node_modules" || strings.HasPrefix(entry.Name(), ".") {
```

### DIFF
--- a/plugins/jsvm/jsvm.go
+++ b/plugins/jsvm/jsvm.go
@@ -440,6 +440,10 @@ func (p *plugin) watchHooks() error {
 	//
 	// @todo replace once recursive watcher is added (https://github.com/fsnotify/fsnotify/issues/18)
 	dirsErr := filepath.WalkDir(watchDir, func(path string, entry fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
 		// ignore hidden directories, node_modules, symlinks, sockets, etc.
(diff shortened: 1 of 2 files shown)


ganigeorgiev:
OK, there is a limit to the AI spam I can handle.

I will review the suggested fix locally but you are not welcome here anymore.

ganigeorgiev:
The suggested fix here is actually incorrect because in case for example of a permission error we don't want to stop the files watcher entirely and instead we need to either skip the file from the watch list or add it to the watch list and do nothing with it _(essentially how it works at the moment)_.

I'll evaluate a little bit later today what issues the current handling can cause and whether ignoring the file entirely would be better option here.

PR #7636, opened 2026-04-11
Fix typos in RecordUpsertPanel.svelte

emileferreira:
This fixes the incorrect infinitive construction in the email confirmation dialogues.


ganigeorgiev:
Thank you but there is no need for this as the UI has been refactored and will be replaced within the next couple weeks (https://github.com/pocketbase/pocketbase/discussions/7612).

PR #7639, opened 2026-04-12
(AI ad)

orbisai0security:
(AI ad)


ganigeorgiev:
There is no vulnerability here.

It is godoc comment example how to set a raw bcrypt hash (the line before it literally explains that...).

Whoever is the dumb operator behind this "AI" - please stop.

